PR #4, opened 2024-08-23
Introduce LICENSE

masutaka:
Introduce `LICENSE`.
We use the Apache-2.0 license.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,178 @@
+Copyright (c) 2024 ROUTE06, Inc.
+
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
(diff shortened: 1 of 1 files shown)


LICENSE
```diff
@@ -0,0 +1,178 @@
+Copyright (c) 2024 ROUTE06, Inc.
+
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
```
toyamarinyon, line 5: How do you renew the year? Is it manual or automatic?

masutaka: @toyamarinyon
Copyright year is the start year. Therefore, it will not be updated.

FYI:
It seems that Rails and curl do not write Copyright year.

* https://github.com/rails/rails/pull/47467
* [Copyright without years \| daniel\.haxx\.se](https://daniel.haxx.se/blog/2023/01/08/copyright-without-years/) (curl article)

It looks like they decided it was unnecessary based on [Berne Convention](https://en.wikipedia.org/wiki/Berne_Convention), which has been ratified by 181 countries. Other organizations are beginning to remove it.

---

The "January 2004" is the date of [Apache-2.0 license](https://www.apache.org/licenses/LICENSE-2.0).

toyamarinyon: Thank you for providing this valuable information. It's very insightful 👍

PR #6, opened 2024-09-12
Add CODEOWNERS

masutaka:
I want to automatically set a reviewer when I request a review via pull request.
If it is the current stage, it looks good to set it automatically.


masutaka:
Thank you for your reviews!

PR #9, opened 2024-10-04
Add "Add assignee to PR" workflow

masutaka:
## Summary

Introduce the reusable workflow [route06/actions/.github/workflows/add_assignee_to_pr.yml](https://github.com/route06/actions/blob/v2/.github/workflows/add_assignee_to_pr.yml)

## Related Issue

Nothing

## Changes

Create `.github/workflows/add_assignee_to_pr.yml`. This workflow was added to [route06inc/template](https://github.com/route06inc/template) by https://github.com/route06inc/template/pull/3

## Testing

✅
It works well.

## Other Information

Nothing

### DIFF
--- a/.github/workflows/add_assignee_to_pr.yml
+++ b/.github/workflows/add_assignee_to_pr.yml
@@ -1,0 +1,11 @@
+name: Add assignee to PR
+
+on:
+  pull_request:
+    types: [opened]
+
+jobs:
+  add-assignee-to-pr:
+    permissions:
+      pull-requests: write
+    uses: route06/actions/.github/workflows/add_assignee_to_pr.yml@v2


masutaka:
@toyamarinyon I just followed the convention of ROUTE06 and did not think too deeply about it.
If there is an advantage, it is that it may be used for data analysis in the future when the number of OSS repositories grows.

masutaka:
@toyamarinyon @shige
I have no strong opinion on this PR. 👀
Toyama-san do not want to increase the number of GitHub Actions, and if there are no other requests, I will close this PR next week.

shige:
@masutaka @toyamarinyon It doesn't matter whether I'm here or not, so let's close it now.

PR #15, opened 2024-10-17
Remove unused default files from next and vercel

shige:
## Summary
Remove unused default files from next and vercel

## Related Issue
None

## Other Information
Testing for vercel deploy


shige:
I have confirmed that the deployment to the Vercel preview environment was successful. 🚀 

image

PR #29, opened 2024-10-21
Fix "Can't resolve 'next/font/local/target.css'"

satococoa:
## Summary
Fix "Can't resolve 'next/font/local/target.css'"

Currently, you will encounter the above error when running `bun dev` and opening `http://localhost:3000` due to the absence of the actual font file. Since Rosart is a proprietary font, I have modified how fonts are handled in the development and preview environments.

## Changes
- Created `prepare-font.js`
- Configured it to run before `bun dev` and build

### DIFF
--- a/app/fonts/index.ts
+++ b/app/fonts/index.ts
@@ -1,7 +1,0 @@
-import { Roboto_Serif } from "next/font/google";
-/**
- * Since the Rosart font is a proprietary font, use an alternative font during OSS development.
- */
-export const rosart = Roboto_Serif({
-	variable: "--font-rosart",
-});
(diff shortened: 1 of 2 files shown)


app/fonts/index.ts
```diff
@@ -1,32 +1,7 @@
...
-			weight: "600",
-			style: "normal",
-		},
-		{
-			path: "./Rosart-Bold.woff2",
-			weight: "700",
-			style: "normal",
-		},
-	],
+import { Roboto_Serif } from "next/font/google";
```
satococoa: I haven't given much thought to using Roboto.
Any font is fine with me.

PR #58, opened 2024-10-29
Implement local environment flagging mechanism

toyamarinyon:
In the local environment, environment variables can be used to mimic the Vercel Toolbar.

### Changes:
- Add utility function to evaluate flags based on local environment
- Replace hardcoded return values with calls to the new function
- This enhancement allows better control of feature flags during development

### DIFF
--- a/flags.ts
+++ b/flags.ts
@@ -1,17 +1,21 @@
 import { unstable_flag as flag } from "@vercel/flags/next";
 
+function takeLocalEnv(localEnvironemntKey: string) {
+	if (process.env.NODE_ENV !== "development") {
+		return false;
+	}
+	if (
+		process.env[localEnvironemntKey] === undefined ||
+		process.env[localEnvironemntKey] === "false"
+	) {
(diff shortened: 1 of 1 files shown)


shige (approved):
LGTM 👍 

---

**OTA**

These feature flags work with the following environments.

- [x] preview
- [x] development

---

**IMO**

I thought that using the `FLAG_` prefix might make the code for environment variables more readable, but I’ll leave it up to you since it's a matter of personal preference.

toyamarinyon:
> I thought that using the `FLAG_` prefix might make the code for environment variables more readable, but I’ll leave it up to you since it's a matter of personal preference.

I understand. In this case, I used the name of the environment variable aligned with the name of the variable in the code, but in the future, if there are any problems, I would be happy to add the suggested Prefix to the form.

https://github.com/route06inc/giselle/blob/a22c35997bd0b63b21e52160383716a7bf544dbd/flags.ts#L15-L65

PR #61, opened 2024-10-29
[For contributors] Enable the feature flag at all times in the development environment

shige:
## Summary
Enable the feature flag at all times in the development environment

## Related Issue
- https://github.com/route06inc/giselle/pull/58

## Changes
Just updateed .env.example

## Testing
Please change the relevant parts of the `.env.sample` file and test it in the development environment.


shige:
Thx!

PR #65, opened 2024-10-30
refactor(graph): Inline debounce logic in GraphProvider

toyamarinyon:
Simplify state management by removing separate useDebounce hook and incorporating debouncing directly in the GraphProvider component. The functionality remains the same but with cleaner code organization.

Key changes:
- Remove standalone useDebounce hook
- Add timeout handling directly in enhancedDispatch
- Clean up timeout on component unmount
- Fix potential memory leak by clearing timeout between state updates

The 500ms debounce delay for graph updates is preserved.

### Verification Steps
The following operations should work without regression:

- [ ] Display Playground.
- [ ] Move the node
- [ ] Wait 1 second
- [ ] Reload
- [ ] The node is displayed at the position where you moved it.


shige (approved):
LGTM 👍

PR #66, opened 2024-10-31
Add updated at column

koji-m:
## Summary
ref. https://github.com/route06inc/giselle/issues/50

In this PR, I will add columns that represent the update datetime (timestamp) of records to several tables in the DB.
Currently, add a column that records record update timestamps only to tables that have record creation timestamps.
(Exclude the `files` and `requests` tables as they are not used.)

## Related Issue
- https://github.com/route06inc/giselle/issues/50

## Changes
- Added the `updatedAt` column to the tables below. (drizzle/schema.ts)
  - agents
  - organizations
  - teams
- Added migration to apply schema changes to DB (migrations/0005_overrated_scarlet_spider.sql)

## Testing
- Create an agent.
  - Verify that new records are created in `agents` and `updated_at` == `created_at` is set.
- Modify agent structure. (ex. add some nodes)
  - Verify that `updated_at` of the modified agent's record has been updated.
- Create an account.
  - Verify that new records are created in `organizations` and `teams` and `updated_at` == `created_at` is set.

## Other Information
- In the column definition of schema.ts, `defaultNow()` is specified, so the initial value of `update_at` for records that already exist are set to the current timestamp. So that, it does not accurately reflect the actual last update time. The alternatives considered are:
  - Same value as `created_at`
    - Like the current timestamp, it does not accurately reflect the actual last update time.
    - After migration, I will need to manually set the `updated_at` value to the `created_at` value.
    - Therefore, there is no reason to update to the same value as `created_at` after changing the schema.
  - NULL
    - This column does not allow nulls.


docs/packages-license.md
```diff
@@ -1,6 +1,6 @@
 # giselle
 
-As of October 29, 2024 10:23am. 926 total
+As of October 31, 2024  8:07am. 926 total
```
toyamarinyon: I think this change is unnecessary.

toyamarinyon (commented):
Thanks, I will check it out! I think it's generally good, but I did comment on one point, so please check it out.

shige (approved):
LGTM 👍

docs/packages-license.md
koji-m: There were no changes to the dependencies, but since bun.lockb was included in the early commit, that file was automatically updated. I reverted it because it is unnecessary. thank you. https://github.com/route06inc/giselle/pull/66/commits/cb4dad8493811afeb6a291f94cba1a82bca15dda

toyamarinyon (approved):
👍

koji-m:
Thank you!

shige:
FYI: @koji-m @toyamarinyon 
I executed the following command! 

```sh
bun drizzle-kit migrate
```

koji-m:
Thank you 🙏

PR #68, opened 2024-11-05
Prototype flow

toyamarinyon:
We have implemented a Run button that can sequentially execute flows created in the Playground (This is a Private Beta feature. It does not affect all users).

Please follow the steps below to verify that the Playground continues to work as before and that flows are executed when pressing the Run button.

### Verification Steps
The following operations should work without regression:

- [ ] Verification of No Regression

    - [ ] 1. Text Generator Node
        - [ ] Place a Text Generator node
        - [ ] Enter text and press Push Value
        - [ ] Verify text generation works
    
    - [ ] 2. Web Search Node
        - [ ] Place a Web Search node
        - [ ] Enter text and press Push Value
        - [ ] Verify search results are generated
    
    - [ ] 3. Prompt Node Source Operations
        - [ ] Add text content and verify text generation
        - [ ] Add file and verify text generation
        - [ ] Add Text Generator artifacts and verify text generation
        - [ ] Add Web Search artifacts and verify text generation
  
    - [ ] Verify that the flow executes sequentially when pressing the Run Button
            
        > [!NOTE]
        > Currently, you need to connect two or more flows, so please connect the flows and verify.
        
        image

        - [ ] 1. Press the Run button
        - [ ] 2. Switch to View screen
        - [ ] 3. Flow executes sequentially
        - [ ] 4. Click tabs on the left side to check execution status of each flow

### DIFF
--- a/app/(playground)/p/[agentId]/beta-proto/artifact/render.tsx
+++ b/app/(playground)/p/[agentId]/beta-proto/artifact/render.tsx
@@ -1,0 +1,27 @@
+import { MarkdownRender } from "../giselle-node/components/panel/markdown-render";
+import type { Artifact } from "./types";
+
+interface ArtifactRenderProps {
+	title: string;
+	content: string;
+}
+export function ArtifactRender(props: ArtifactRenderProps) {
+	return (
+		<div>
+			<div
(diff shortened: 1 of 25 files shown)


shige (approved):
LGTM! 🆙

PR #70, opened 2024-11-05
Extend serverless function timeout to 5 minutes

toyamarinyon:
Replace global Vercel function timeout config with per-route maxDuration 
setting for the playground route. Increases timeout from 60s to 300s to 
accommodate longer-running agent operations.

> [!NOTE]
> This change only affects the /p/[agentId] playground route where
extended processing time may be needed.


toyamarinyon:
I've created a new pull request (#74)

PR #71, opened 2024-11-05
Replace Trigger.dev with Vercel Functions

toyamarinyon:
- Remove @trigger.dev/sdk dependency and configuration
- Delete task runners for account initialization and request handling
- Clean up unused workflow orchestration code

At this time, Vercel Function is sufficient, and the Trigger.dev function is unnecessary and has many dependencies that often conflict with the update process, so it will be removed once.


shige (approved):
LGTM 🚿

PR #73, opened 2024-11-05
Use next.config.ts instead of next.config.mjs

toyamarinyon:
Next.js 15 introduced native support for TypeScript configuration files. 
https://nextjs.org/blog/next-15#support-for-nextconfigts

Allowing us to use next.config.ts instead of next.config.mjs.


shige (approved):
LGTM 🆙

PR #74, opened 2024-11-05
feat: Set 5-minute timeout for Agent playground page

toyamarinyon:
Now that we know how to set maxDuration on a page-by-page basis, we have changed from using vercel.json to specify it for the entire project. 
https://nextjs.org/docs/app/api-reference/file-conventions/route-segment-config#maxduration

Also, when executing the Run function, MaxDuration is now set to the maximum of 5 minutes, since the previous 1 minute may be too short.


shige (approved):
LGTM! ⏲️

PR #75, opened 2024-11-05
fix: Update SVG attributes to React camelCase format

toyamarinyon:
- Change clip-path to clipPath
- Change fill-rule to fillRule
- Change clip-rule to clipRule


shige (approved):
LGTM 👍

PR #77, opened 2024-11-06
Add model selection option to Text Generator node with Claude support

toyamarinyon:
> [!NOTE]
> This modifications are a Private Beta feature. It does not affect all users

- Added ability to select models in Text Generator node when chooseModel flag is enabled
- Added Claude (Anthropic) as a model option when anthropic flag is enabled

https://github.com/user-attachments/assets/61c38b9f-e820-44d3-8532-8e73e0780995


services/agents/nodes/classes/text/build-template.ts
```diff
@@ -9,7 +9,7 @@ import {
 	requestPortMessages,
 	requests,
 } from "@/drizzle";
-import { logger } from "@trigger.dev/sdk/v3";
+// import { logger } from "@trigger.dev/sdk/v3";
```
shige, line 12: nits: I thought it might be fine to delete all the commented-out lines in this file.

shige (approved):
LGTM 👍 

---

> This modifications are a Private Beta feature. It does not affect all users

Anthropic API is available to all users from today, so I don't think there's any need to set the feature flag.

- see: https://giselle-public.s3.amazonaws.com/legal/terms/2024-10-30-Giselle-Terms-of-Service.pdf

PR #78, opened 2024-11-07
feat(llm): Add configurable model provider support

toyamarinyon:
In #77 we supports configurable model for Flow. This one also supports for Playground.

Add ModelConfiguration type and utilities to support different LLM providers (OpenAI, Anthropic) with customizable parameters like temperature and topP. Replace hardcoded GPT-4 model with dynamic model selection based on node properties.

- Introduce ModelProvider type and ModelConfiguration interface
- Add resolveModelConfiguration utility for model selection
- Update generateArtifactStream to use configurable models
- Set default fallback configuration to gpt-4o

Related to text generation workflow improvements


shige (approved):
LGTM! ⚙️

PR #82, opened 2024-11-08
fix(a11y): Add required DialogTitle to improve dialog accessibility

toyamarinyon:
Add missing DialogTitle components to dialog implementations to ensure proper accessibility for screen reader users, following Radix UI Dialog requirements.

- Replace div headers with DialogTitle in artifact-block
- Replace div headers in websearch-block
- Maintain existing styling while improving ARIA compliance

Fixes accessibility issues with Dialog component implementation per Radix UI documentation requirements.


shige (approved):
LGTM 👍

Issue #84, opened 2024-11-08
Telemetry data doesn't arrive at observability backend


Rindrics:
### Self Checks

- [X] This is only for bug report, if you would like to ask a question, please head to [Discussions](../discussions/categories/q-a).
- [X] I have searched for existing issues [search for existing issues](../issues), including closed ones.

### Cloud or Self Hosted

Cloud

### Steps to reproduce

1. visit https://studio.giselles.ai/agents
1. select an agent
1. push query to agent
1. (private environment) visit observability backend
1. (private environment) search for the corresponding telemetry data

### Expected Behavior

telemetry data, such as token consumed, is arrived at observability backend

### Actual Behavior

expected data does not arrive

### Additional Context

core problem:
- cycle of telemetry exporting by OTEL exporter is independent of lifecycle of Vercel Functions
- if core process of server action completes earlier than exporters' next exporting interval, Vercel Function finishes with telemetry data unexported

current Giselles.ai
- uses [Sentry](https://sentry.io) as error tracker
- overrides Vercel Function max duration with [30 seconds](https://github.com/giselles-ai/giselle/blob/170274e947dfe7d7ecbea969e7dfc5319e12fa2d/app/(playground)/p/%5BagentId%5D/page.tsx#L27)
- has Vercel Collector [on `localhost:4318`](https://github.com/vercel/otel/blob/087601ae585cb116bb2b46c211d014520de76c71/packages/otel/src/exporters/trace-config.ts#L4) but don't use it because [it only supports New Relic and Datadog at this point](https://vercel.com/docs/observability/otel-overview#:~:text=or%20Edge%20Functions.-,Prerequisites,teams%20on%20Pro%20and%20Enterprise%20plans%20by%20following%20the%20steps%20below,-Get%20started)

PR #85, opened 2024-11-08
Keep Vercel Function awake during telemetry exporting interval

Rindrics:
## Summary

Ensure telemetry data sent to observability backend by keeping Vercel Function awake during telemetry exporting interval

## Related Issue

- this PR resolves https://github.com/giselles-ai/giselle/issues/84

## Changes

- use `waitUntil()` to wait for 15 seconds after object generation finished

## Testing

I confirmed telemetry arrived at observability backend (private environment)

## Other Information

this change does not affect response time of core use case because `waitUntil()` does not block response

ref: https://vercel.com/docs/functions/functions-api-reference#waituntil


sentry.server.config.ts
```diff
@@ -13,4 +13,5 @@ Sentry.init({
 	// Setting this option to true will print useful information to the console while you're setting up Sentry.
 	debug: false,
 	enabled: process.env.NEXT_PUBLIC_VERCEL_ENV !== undefined,
+	skipOpenTelemetrySetup: true,
```
rindrics, line 16: needed to avoid setup conflict

instrumentation.ts
```diff
@@ -2,6 +2,7 @@ import * as Sentry from "@sentry/nextjs";
 
 export async function register() {
 	if (process.env.NEXT_RUNTIME === "nodejs") {
+		await import("./instrumentation.node");
```
rindrics, line 5: to initialize OTEL clients on app startup

instrumentation.node.ts
```diff
@@ -1,27 +1,13 @@
...
+} from "@/lib/opentelemetry";
+import { registerOTel } from "@vercel/otel";
 
-// Configure OTLP metrics exporter
-const otlpMetricsExporter = new OTLPMetricExporter({
-	url: "https://ingest.us.signoz.cloud:443/v1/metrics",
-	headers: {
-		"signoz-access-token": process.env.SIGNOZ_INGESTION_TOKEN,
-	},
+registerOTel({
```
rindrics, line 8: use `registerOTel()` to simplify SDK configuration

without explicit configuration of OTEL Resource, we can gather useful info such as:
- Vercel build ID
- service version
- (and more)

lib/opentelemetry/trace.ts
```diff
@@ -0,0 +1,3 @@
+import { NoopSpanProcessor } from "@opentelemetry/sdk-trace-base";
+
+export const noopSpanProcessor = new NoopSpanProcessor();
```
rindrics, line 3: Although we don't collect traces currently, we need to pass this blank processor to `registerOTel()`
- to avoid [auto initialization of span processor inside `registerOTel()`](https://github.com/vercel/otel/blob/087601ae585cb116bb2b46c211d014520de76c71/packages/otel/src/sdk.ts#L409-L415)

package.json
```diff
@@ -20,6 +20,7 @@
 		"@next/env": "14.2.5",
 		"@next/third-parties": "^14.2.13",
 		"@opentelemetry/api": "^1.9.0",
+		"@opentelemetry/exporter-logs-otlp-http": "^0.54.2",
```
toyamarinyon: Set the version exactly
```suggestion
		"@opentelemetry/exporter-logs-otlp-http": "0.54.2",
```

package.json
```diff
@@ -43,6 +44,7 @@
 		"@supabase/supabase-js": "2.45.0",
 		"@vercel/blob": "0.23.4",
 		"@vercel/flags": "2.6.1",
+		"@vercel/functions": "^1.5.0",
```
toyamarinyon: Set the version exactly
```suggestion
		"@vercel/functions": "1.5.0",
```

app/(playground)/p/[agentId]/beta-proto/artifact/server-actions.ts
```diff
@@ -80,6 +81,16 @@ ${sourcesToText(sources)}
 					output: result,
 				});
 				await lf.shutdownAsync();
+				waitUntil(
+					new Promise((resolve) =>
+						setTimeout(
+							resolve,
+							Number.parseInt(
+								process.env.VERCEL_FUNCTION_MAX_DURATION ?? "300",
+							),
```
Comment (toyamarinyon) [inline app/(playground)/p/[agentId]/beta-proto/artifact/server-actions.ts]: I think it should be aligned with OTEL's exportInterval instead of Vercel Function's execution time limit.

```suggestion
							Number.parseInt(
								process.env.OTEL_EXPORT_INTERVAL_MILLIS?? "1000",
							),
```

And modify `lib/opentelemetry/metric.ts`

```ts
export const metricReader = new PeriodicExportingMetricReader({
	exporter: metricExporter,
	exportIntervalMillis: Number.parseInt(
								process.env.OTEL_EXPORT_INTERVAL_MILLIS?? "1000",
							),
	exportTimeoutMillis: 900, // retries exporting if timeout
});
```

rindrics:
thanks, I applied the suggested changes @toyamarinyon

toyamarinyon (commented):
Thank you very much. Let's see how it goes with this.

shige (approved):
LGTM 👍

rindrics:
thx!

PR #88, opened 2024-11-12
Add structured logging system with configurable log levels

### LINKED ISSUE #87: Enable controlling log level
## Task description

Implement a leveled logging system that enables structured output for debugging

## Background

We need to strike a balance between two competing needs:
- Including detailed debug information for development and troubleshooting
- Maintaining clean and focused console output for production use

Key considerations:
- Development environments need verbose logging for debugging
- Production environments should show only essential information

## Additional notes

- It would be better if the logger tool have structured-logging function

Rindrics:
## Summary

This PR enables:
- configurable log levels
- structured logging

## Related Issue

- close #87

## Changes

- introduced dependency to [pino](https://getpino.io/)

## Testing

I confirmed debug log only when I ran `bun run debug` 
```
{"level":20,"time":1731382500233,"pid":11887,"hostname":"my-machine","tokenConsumed":243,"msg":"response obtained"}
```

## Other Information

- The debug log I included in this PR won't show up in the production log because the default log level was set to `info`


toyamarinyon (commented):
Please tell me why you chose pino as logger library. Is it necessary for our purpose? Can't we implement it ourselves?

rindrics:
@toyamarinyon 

In addition to structured-logging function, we need the reduct (mask) function of pino because we are planning to inject logs to BI platform.

Reduct function allows easy masking of sensitive info [by JSON path](https://github.com/pinojs/pino/blob/9047cd9492961ee9b8f3f17085a7a9a87a88cede/docs/redaction.md).

When we introduce such function, adopting well-tested preexistent package helps us focus on developing core function of our app

Pino is simple enough for our purpose compared to other logger packages such as winston and bunyan.

We should consider the recent increasing trend of pino community because it affects the size of available knowledge base on the internet.
ref: https://npmtrends.com/bunyan-vs-pino-vs-winston

toyamarinyon (approved):
I see, it is necessary for reductions. I think it is reasonable to use a library for this purpose. Among the logging libraries, pino is a good one, in my opinion, so it is good.

package.json
```diff
@@ -5,6 +5,7 @@
 	"scripts": {
 		"predev": "bun prepare-font.js",
 		"dev": "SENTRY_SUPPRESS_TURBOPACK_WARNING=1 next dev --turbo",
+		"debug": "LOGLEVEL=debug bun run dev",
```
shige:
```suggestion
		"dev-debug": "LOGLEVEL=debug bun run dev",
```

shige: 🍠 imo

I thought it might be better to use `dev-debug` explicitly, as just `debug` makes it hard to distinguish what kind of `debug` it is.

example: https://github.com/vercel/next.js/blob/d6dd69f210eb36360e6b4544a5f7a6918e8bdb63/package.json#L18

shige (approved):
LGTM 🍨

package.json
rindrics: thanks!

the suggestion applied: [3b9c098](https://github.com/giselles-ai/giselle/pull/88/commits/3b9c098ad28c20dbc97323120ba294c38d41be46)

PR #91, opened 2024-11-12
feat(webhooks): Switch from issues to issue_comment handler

toyamarinyon:
Add integration-based comment handling for issues with "/giselle" command

- Replace issues webhook with issue_comment webhook handler
- Add integration retrieval logic for repository and call sign
- Implement command pattern using "/giselle" prefix
- Support multiple integrations with parallel processing

The new handler allows for more flexible interaction by responding to comments rather than just issue events, enabling command-based workflows through the "/giselle" prefix.

### DIFF
--- a/app/webhooks/github/route.ts
+++ b/app/webhooks/github/route.ts
@@ -9,7 +9,7 @@ import { captureException } from "@sentry/nextjs";
 import type { NextRequest } from "next/server";
 
 function setupHandlers() {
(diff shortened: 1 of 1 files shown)


toyamarinyon:
I checked that it works.

https://github.com/toyamarinyon/ai-sdk-langfuse/issues/1#issuecomment-2470409682

image

Issue #53, opened 2024-10-28
Implement GitHub social login


shige:
## Task description
Implement GitHub social login functionality using Supabase as outlined in the Supabase Docs.

## Background
Integrating GitHub login allows users to authenticate using their GitHub credentials, providing a seamless user experience and reducing the need for password management. This is particularly useful for developers, who are likely to have GitHub accounts. Supabase provides a straightforward guide for implementing social logins, ensuring a secure and efficient authentication process.

## TODO (Optional)

- [x] Set up a Supabase project and configure GitHub as a provider in the Supabase Dashboard.
- [x] Update the Giselle application to include the option for users to log in using their GitHub account.
- [x] Test the GitHub login feature.

## Additional notes (Optional)

- Ensure you have a GitHub account with developer access to create and configure OAuth applications.
- Ensure that environment variables for client ID and secret are securely managed and not hard-coded into the client-side code.
- Reference the [Supabase Docs on GitHub social login](https://supabase.com/docs/guides/auth/social-login/auth-github) for detailed setup instructions.
- Consider the privacy implications and inform users about the data sharing involved with using third-party login services.

shige:
📆 We've updated the Terms of Service, which will be released after the Effective Date: November 08, 2024.

---

https://giselles.ai/legal/terms

> Effective Date: November 08, 2024

satococoa:
Successfully released 🚀

All features have been deployed and verified in production environment.

PR #93, opened 2024-11-12
Provide google login

Rindrics:
## Summary

Enables user to login using Google account

## Related Issue

- #53 

## Changes

- add "Google" type social login button
- unify provider detection logic between GitHub and Google

## Testing

- [x] "Sign in with Google" button is absent by default
- [x] "Sign in with Google" button is available with its feature flag on
- [x] "Sign in with Google" login works
- [x] preexistent "Sign in with GitHub" works

### DIFF
--- a/app/(auth)/actions.ts
+++ b/app/(auth)/actions.ts
@@ -4,11 +4,12 @@ import { getAuthCallbackUrl } from "@/app/(auth)/lib";
 import { createClient } from "@/lib/supabase";
 import { redirect } from "next/navigation";
 
-export async function authorizeGitHub() {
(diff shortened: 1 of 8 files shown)


rindrics:
> ### Testing
> - [x] "Sign in with Google" button is absent by default
> - [x] "Sign in with Google" button is available with its feature flag on
> - [x] "Sign in with Google" login works
> - [x] preexistent "Sign in with GitHub" works

Checks are all met, but I'll re-create another PR with same content because Vercel preview env with "branch" type domain (private environment) does not work with some reason.

PR #95, opened 2024-11-13
Restrict multiple github account linking

satococoa:
## Summary
The Supabase Manual linking feature allows users to link multiple identities from the same provider.
However it introduces additional complexity, and not suitable for most use cases.

I've added validation to check if the user already has a GitHub identity before executing the link identity process.

https://supabase.com/docs/guides/auth/auth-identity-linking#manual-linking-beta


toyamarinyon (approved):

shige (approved):
LGTM 🛡️

PR #96, opened 2024-11-13
Add retry mechanism for getUser function

satococoa:
Supabase Auth sometimes may return AuthRetryableFetchError.

Introduce a `withRetry` utility function to handle retries for asynchronous operations. Update the `getUser` function to utilize this utility, allowing it to retry fetching the user on specific authentication errors.

### DIFF
--- a/lib/supabase/get-user.ts
+++ b/lib/supabase/get-user.ts
@@ -26,6 +26,7 @@ export const getUser = async () => {
 		return data.user;
 	};
 
+	// No specific reason
 	const retryCount = 5;
 	const user = await withRetry(getUserFunc, {
 		retries: retryCount,
--- a/lib/supabase/get-user.ts
+++ b/lib/supabase/get-user.ts
@@ -26,10 +26,8 @@ export const getUser = async () => {
 		return data.user;
 	};
 
-	// No specific reason
-	const retryCount = 5;
 	const user = await withRetry(getUserFunc, {
-		retries: retryCount,
+		useExponentialBackoff: true,
 		onRetry: (retryCount, error) => {
 			console.error(`getUser failed, retrying (${retryCount})`, error);
 		},
--- a/lib/supabase/get-user.ts
+++ b/lib/supabase/get-user.ts
@@ -1,3 +1,5 @@
+import { isAuthRetryableFetchError } from "@supabase/supabase-js";
+import { withRetry } from "../utils";
 import { createClient } from "./server";
 
 /**
@@ -12,12 +14,29 @@ import { createClient } from "./server";
 export const getUser = async () => {
 	const supabase = await createClient();
 
-	const { data, error } = await supabase.auth.getUser();
-	if (error != null) {
-		throw error;
-	}
-	if (data.user == null) {
-		throw new Error("No user returned");
-	}
-	return data.user;
+	const getUserFunc = async () => {
+		const { data, error } = await supabase.auth.getUser();
+
+		if (error != null) {
+			throw error;
+		}
+		if (data.user == null) {
+			throw new Error("No user returned");
+		}
+		return data.user;
+	};
+
+	const retryCount = 5;
+	const user = await withRetry(getUserFunc, {
+		retries: retryCount,
+		onRetry: (retryCount, error) => {
+			console.error(`getUser failed, retrying (${retryCount})`, error);
+		},
+		shouldAbort: (error) => {
+			// retry if the error is a retryable fetch error
+			return !isAuthRetryableFetchError(error);
+		},
+	});
+
+	return user;
 };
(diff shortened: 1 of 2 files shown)


lib/supabase/get-user.ts
```diff
@@ -12,12 +14,29 @@ import { createClient } from "./server";
...
+		if (error != null) {
+			throw error;
+		}
+		if (data.user == null) {
+			throw new Error("No user returned");
+		}
+		return data.user;
+	};
+
+	const retryCount = 5;
```
shige: nits:
If there is any context behind setting the value to 5, I would appreciate it if you could include it in a comment. 
Even if it's just a fixed value, I would appreciate it if you could indicate that as well.

shige (approved):
LGTM! 🔁

lib/supabase/get-user.ts
satococoa: There is no specific reason.
So I'll add comment to indicate that.

lib/utils.ts
toyamarinyon, line 1: If you use the retry function, it's better to implement exponential backoff mechanism. For example:

```ts
/**
 * Configuration options for the retry mechanism
 */
export interface RetryOptions {
  /** Maximum number of retry attempts */
  maxRetries: number;
  /** Initial delay between retries in milliseconds */
  initialDelay?: number;
  /** Maximum delay between retries in milliseconds */
  maxDelay?: number;
  /** Backoff factor for exponential delay calculation */
  backoffFactor?: number;
  /** Callback function executed before each retry attempt */
  onRetry?: (retryCount: number, error: unknown) => void | Promise<void>;
  /** Function to determine if retries should be aborted */
  shouldAbort?: (error: unknown) => boolean;
  /** Whether to use exponential backoff between retries */
  useExponentialBackoff?: boolean;
}

/**
 * Error thrown when maximum retry attempts are exceeded
 */
export class MaxRetriesExceededError extends Error {
  constructor(public readonly errors: unknown[]) {
    super("Maximum retry attempts exceeded");
    this.name = "MaxRetriesExceededError";
    Object.setPrototypeOf(this, MaxRetriesExceededError.prototype);
  }
}

/**
 * Utility function to delay execution
 */
const delay = (ms: number): Promise<void> =>
  new Promise(resolve => setTimeout(resolve, ms));

/**
 * Calculate delay time using exponential backoff
 */
const calculateDelay = (
  attempt: number,
  initialDelay: number,
  maxDelay: number,
  backoffFactor: number
): number => {
  const exponentialDelay = initialDelay * Math.pow(backoffFactor, attempt);
  return Math.min(exponentialDelay, maxDelay);
};

/**
 * Executes an asynchronous function with configurable retry behavior
 * 
 * @param fn - The async function to execute
 * @param options - Retry configuration options
 * @returns Promise resolving to the function's return value
 * @throws MaxRetriesExceededError if max retries are reached
 */
export async function withRetry<T>(
  fn: () => Promise<T>,
  options: Partial<RetryOptions> = {}
): Promise<T> {
  const {
    maxRetries = 3,
    initialDelay = 1000,
    maxDelay = 30000,
    backoffFactor = 2,
    onRetry = async () => {},
    shouldAbort = () => false,
    useExponentialBackoff = true
  } = options;

  const errors: unknown[] = [];

  for (let attempt = 0; attempt < maxRetries; attempt++) {
    try {
      return await fn();
    } catch (error: unknown) {
      if (shouldAbort(error)) {
        throw error;
      }

      errors.push(error);
      
      // Don't delay on the last attempt
      if (attempt < maxRetries - 1) {
        await onRetry(attempt, error);
        
        if (useExponentialBackoff) {
          const delayTime = calculateDelay(
            attempt,
            initialDelay,
            maxDelay,
            backoffFactor
          );
          await delay(delayTime);
        } else {
          await delay(initialDelay);
        }
      }
    }
  }

  throw new MaxRetriesExceededError(errors);
}
```

shige (approved):
Thanks 👍

toyamarinyon (commented):
Is there any reasonable chance that retrying the getAuth function will resolve the AuthRetryableFetchError?

I was wondering if I could try to find the cause of the AuthRetryableFetchError first.

satococoa:
@toyamarinyon 

> Is there any reasonable chance that retrying the getAuth function will resolve the AuthRetryableFetchError?

`AuthRetryableFetchError` is thrown from these functions.

https://github.com/supabase/auth-js/blob/67d3e865101f8e9b7c8e5332c622722c67f10d33/src/lib/fetch.ts#L40-L48
https://github.com/supabase/auth-js/blob/67d3e865101f8e9b7c8e5332c622722c67f10d33/src/lib/fetch.ts#L171-L192

It appears to be caused by a network issue or a problem with the Supabase server. 
Therefore, the best course of action is to retry the request.

lib/utils.ts
satococoa: I agree 👍 
I'll implement exponential backoff feature.

satococoa: implemented
[8c33a5a](https://github.com/giselles-ai/giselle/pull/96/commits/8c33a5a65db8cb4088b4e98787e083e74f45adbc)

toyamarinyon (approved):
Thanks!

PR #101, opened 2024-11-13
Respect given x-forwarded-* headers

satococoa:
## Summary

- Handle x-forwarded-host and x-forwarded-proto headers to correctly redirect requests.
  - With this, you can successfully use OAuth login using tunneling software (e.g., Cloudflare Tunnel).
- Ensure fallback to https if x-forwarded-proto is not set.

[FYI] `next dev` requests also have `x-forwarded-host` and `x-forwarded-proto` headers. I don't know the internal details, but that was a surprise to me.


toyamarinyon (approved):

shige (approved):
LGTM! ✖️

PR #105, opened 2024-11-13
style(tabs): Add backdrop blur effect and adjust background color

toyamarinyon:
Enhances visual styling of tab content panel by adding blur effect and slightly adjusting the background color for better depth perception.

Changes:
- Add backdrop-blur-[16px] for frosted glass effect
- Update background HSL value for subtle color refinement


shige (approved):
LGTM 👍

PR #108, opened 2024-11-18
feat: make call sign input controlled and update preview dynamically

toyamarinyon:
Updates the GitHub integration form to use controlled input for the call sign field, showing real-time updates in the usage example preview. Previously used uncontrolled input with defaultValue.


shige (approved):
LGTM 👍

PR #112, opened 2024-11-20
Add event capturing into PostHog

koji-m:
## Summary
Added feature to allow PostHog to capture access events for user behavior analysis.
This feature can be enabled using environment variables.

## Changes
- Added PostHog initializing component to `app/providers.tsx` .
- Added event capturing component to `app/posthog-page-view.tsx` .
- Inserted components implemented above into `app/layout.tsx` .

### Normal case - Enable event capturing
- Set environment variables
  - `NEXT_PUBLIC_ENABLE_POSTHOG` : true (enables event capturing)
  - `NEXT_PUBLIC_POSTHOG_HOST` : PostHog test project url
  - `NEXT_PUBLIC_POSTHOG_KEY` : PostHog test project key
- Start up development environment ( `bun dev` ) and access some pages.
- Result: Verified that the event was recorded in PostHog.

###  Normal case - Disable event capturing
- Set environment variables
  - `NEXT_PUBLIC_ENABLE_POSTHOG` : false or remove definition (disables event capturing)
  - `NEXT_PUBLIC_POSTHOG_HOST` : PostHog test project url
  - `NEXT_PUBLIC_POSTHOG_KEY` : PostHog test project key
- Start up development environment ( `bun dev` ) and access some pages.
- Result: Verified that the event was not recorded in PostHog.

### Error case
- Set environment variables
  - `NEXT_PUBLIC_ENABLE_POSTHOG` : true (enables event capturing)
  - `NEXT_PUBLIC_POSTHOG_HOST` : Wrong PostHog test project url or remove definition
  - `NEXT_PUBLIC_POSTHOG_KEY` : Wrong PostHog test project key or remove definition
- Start up development environment ( `bun dev` ) and access some pages.
- Result: Although an error is displayed on the browser console (not in app), confirmed that the application runs normally.

## Other Information
- I referred to [PostHog's Next.js implementation guide](https://posthog.com/docs/libraries/next-js).
- If an error occurs related to event capture, the error is reported to Sentry without affecting the operation of the application.


app/Tracking.tsx
toyamarinyon, line 1: File names in this project are aligned in lower case, so please use lower case.

app/providers.tsx
```diff
@@ -0,0 +1,43 @@
...
+import { useEffect } from "react";
+
+export function Analytics({
+	children,
+}: {
+	children: React.ReactNode;
+}) {
+	if (process.env.NEXT_PUBLIC_ENABLE_ANALYTICS !== "true") {
+		return <>{children}</>;
+	}
```
toyamarinyon: This code runs before the useEffect, but after the full function component body. This means even when analytics are disabled, the component will still try to use the PostHogProvider, which could cause errors since posthog won't be initialized.

https://react.dev/reference/rules/rules-of-hooks

> [!NOTE]
> I was expecting an error with biome, but was surprised that it didn't 🤔 

A better structure would be:

```tsx
export function Analytics({ children }: { children: React.ReactNode }) {
    if (process.env.NEXT_PUBLIC_ENABLE_ANALYTICS !== "true") {
        return <>{children}</>;
    }

    return (
        <AnalyticsEnabled>
            {children}
        </AnalyticsEnabled>
    );
}

function AnalyticsEnabled({ children }: { children: React.ReactNode }) {
    useEffect(() => {
        try {
            // ... initialization code ...
        } catch (error) {
            // ... error handling ...
        }
    }, []);

    return <PostHogProvider client={posthog}>{children}</PostHogProvider>;
}
```

app/providers.tsx
```diff
@@ -0,0 +1,43 @@
+"use client";
+
+import * as Sentry from "@sentry/nextjs";
+import posthog from "posthog-js";
+import { PostHogProvider } from "posthog-js/react";
+import { useEffect } from "react";
+
+export function Analytics({
```
toyamarinyon: Please tell me why do you've changed the name from example.

https://posthog.com/docs/libraries/next-js

image

package.json
```diff
@@ -71,6 +71,7 @@
 		"next-themes": "0.3.0",
 		"openai": "4.64.0",
 		"pino": "9.5.0",
+		"posthog-js": "^1.187.1",
```
toyamarinyon:
```suggestion
		"posthog-js": "1.190.2",
```

toyamarinyon (commented):
Thanks! I've commented!

app/providers.tsx
koji-m: This is a remnant of creating a nested function that included PHProvider() in the early implementation stage. For now, PHProvider is appropriate as it does not need to be this name.

package.json
koji-m: updated https://github.com/giselles-ai/giselle/pull/112/commits/1320ed7603da02ecee1ae237ded751f52fe3f45a

app/providers.tsx
koji-m: Thank you. Fixed https://github.com/giselles-ai/giselle/pull/112/commits/a963775105af8f4b444ee24515f9549dd1531843

app/Tracking.tsx
koji-m: fixed https://github.com/giselles-ai/giselle/pull/112/commits/f892d19ccce61847311e29ab2284322848b56125

app/tracking.tsx
toyamarinyon, line 1: Please tell me why do you've changed the name from example.

image

toyamarinyon (commented):
Thanks for the change. 

Please check my additional comments.

app/tracking.tsx
koji-m: Same as https://github.com/giselles-ai/giselle/pull/112#discussion_r1860595846 , reverted to the same name as example. https://github.com/giselles-ai/giselle/pull/112/commits/f12aee246fd82bcecdfacfd5c30ba0c7ca5a460e

> This is a remnant of creating a nested function that included PHProvider() in the early implementation stage. For now, PHProvider is appropriate as it does not need to be this name.

toyamarinyon (approved):
Thanks!

shige (approved):
LGTM! 🦔

koji-m:
thank you!

PR #113, opened 2024-11-21
fix Langfuse model id

koji-m:
## Summary
When using Claude 3.5 Sonnet, Langfuse cannot calculate the number of tokens etc.
The model id specified for Langfuse's generation() is incorrect, so correct it so that it is an appropriate model id.

## Changes
- Added a function `langfuseModel()` to convert Giselle model id to Langfuse model id.
- Pass the model id output by that function to Langfuse's generation().

## Testing
- Confirmed that all models work in the local dev environment.
  - gpt-4o
  - gpt-4o-mini
  - claude-3.5-sonnet

Test result displayed in Langfuse.

スクリーンショット 2024-11-21 17 27 03


toyamarinyon (approved):
Thanks!!!!

PR #117, opened 2024-11-22
Create `security.txt` and add a GH Actions workflow to remind you of its updates

masutaka:
## Summary

Introduce `security.txt` defined in [RFC 9116](https://www.rfc-editor.org/rfc/rfc9116) as https://studio.giselles.ai/.well-known/security.txt.

Since `security.txt` needs to be updated periodically, I also created a GitHub Actions workflow to automatically create a reminder issue.

## Testing

I set the Expires in `security.txt` to the end of this year, 2024, so that an update reminder issue will be created on 12/1/2024, and that will be the test.

## Other Information

Please let me know if there are any settings needed on vercel.

### DIFF
--- a/.github/workflows/create_issue_security_txt.yml
+++ b/.github/workflows/create_issue_security_txt.yml
@@ -1,0 +1,27 @@
+name: Create an issue to remind updating of security.txt
+
+on:
+  workflow_dispatch:
+  schedule:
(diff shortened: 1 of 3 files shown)


shige:
📝 
I will modify this PR so that the security.txt is applied only to the cloud version.

toyamarinyon (approved):

shige (approved):
LGTM 👍

PR #118, opened 2024-11-22
add error handling to text generation

koji-m:
## Summary
Added temporal error handling for LLM API calls in text generation node processing.
This is a temporary solution because there is a slightly high possibility that an Overloaded error will occur when calling Claude 3.5 Sonnet's API.
Web scraping node uses gpt, so will not support it at this time.

## Changes
- Catch an exception when calling streamObject().
- Display error messages in the right output panel.

## Testing
I confirmed that setting an incorrect value to the LLM API key causes an error message to be output to the right panel.


shige (approved):
LGTM! 🤾

toyamarinyon (approved):

PR #124, opened 2024-11-26
Enable to identify internal team

satococoa:
## Summary
We need to identify teams created by `@route06.co.jp` users for internal dog-fooding purposes.

e.g., Unlimited agent time charge without subscriptions

## Changes
- Added `teams.type` column. Available values are `customer`, `internal`. (defaults to `customer`)
- Added migration script
  - Fetches Supabase user data and checks email domain
- Sets `teams.type` to `internal` when creating a user's default team which user is an internal user (`@route06.co.jp`)

## TODO
- [x] Apply schema migration to Preview environment
- [x] Execute data migration script in Preview environment
- [ ] Apply schema migration to Production environment
- [ ] Execute data migration script in Production environment

## Additional Notes
Initially considered adding `teams.created_by_user_db_id` column, but this approach would require frequent Supabase user data fetching since Giselle doesn't store user emails locally.


scripts/20241126-db-patch-teams-is-internal-team.ts
```diff
@@ -0,0 +1,66 @@
+import { db, supabaseUserMappings, teamMemberships, teams } from "@/drizzle";
+import { createClient } from "@supabase/supabase-js";
+import { eq, inArray } from "drizzle-orm";
+
+console.log("Updating teams to set 'isInternalTeam' property on nodes...");
+const supabaseUrl = process.env.NEXT_PUBLIC_SUPABASE_URL;
+if (!supabaseUrl) {
+	throw new Error("Missing NEXT_PUBLIC_SUPABASE_URL");
+}
+const supabaseServiceRoleKey = process.env.SUPABASE_SERVICE_ROLE_KEY;
```
satococoa, line 10: This key is available at Supabase Dashboard.

scripts/20241126-db-patch-teams-is-internal-team.ts
```diff
@@ -0,0 +1,66 @@
...
+		supabaseUserId: supabaseUserMappings.supabaseUserId,
+	})
+	.from(teams)
+	.innerJoin(teamMemberships, eq(teams.dbId, teamMemberships.teamDbId))
+	.innerJoin(
+		supabaseUserMappings,
+		eq(teamMemberships.userDbId, supabaseUserMappings.userDbId),
+	);
+
+const { data, error } = await supabase.auth.admin.listUsers({
```
satococoa, line 34: We must use `auth.admin` to fetch arbitrary users.

drizzle/schema.ts
```diff
@@ -74,6 +74,7 @@ export const teams = pgTable("teams", {
 		.defaultNow()
 		.notNull()
 		.$onUpdate(() => new Date()),
+	isInternalTeam: boolean("is_internal_team").notNull().default(false),
```
shige: @satococoa 
If we use a boolean like `isInternalTeam`, there is a concern that adding more team types for other purposes in the future could lead to an increase in columns and a combinatorial explosion. 
The term "combinatorial explosion" might be an exaggeration, but having to implement logic to handle combinations of multiple booleans could create a situation prone to bugs.

From my past experience, dealing with such changes was quite challenging, so since this change involves modifying the database table, how about using an enum from the beginning instead of a boolean?

satococoa: @shige 

> The term "combinatorial explosion" might be an exaggeration, but having to implement logic to handle combinations of multiple booleans could create a situation prone to bugs.

Yes, I think so.

> how about using an enum from the beginning instead of a boolean?

Agreed.

I'll make the changes following your advice.
Thank you!

satococoa:
I'll apply migration for Preview environment after code reviews.

shige (approved):
LGTM! 🏷️

drizzle/schema.ts
```diff
@@ -66,6 +66,7 @@ export const subscriptions = pgTable("subscriptions", {
 	trialEnd: timestamp("trial_end"),
 });
 
+type TeamType = "standard" | "internal";
```
toyamarinyon:
```suggestion
type TeamType = "customer" | "internal";
```

I'd like to propose changing the team type from `standard` to `customer`.

Reasons:
- `customer` more clearly conveys the meaning of "customer team", improving code readability
- `standard` could be confused with plan types (e.g., standard plan)
- Looking at future extensibility, naming based on organization types (customer/internal/partner) would be more consistent when we add new team types

Let me know what you think about this suggestion.

satococoa: Thank you. I think `customer` is better too 👍

satococoa: fixed [aa9c5f3](https://github.com/giselles-ai/giselle/pull/124/commits/aa9c5f3f3d7af493507a6cdf7e07b3d47cfb9e99)

toyamarinyon (approved):

satococoa:
I'll run these on preview, now.

- [ ] Apply schema migration to Preview environment
- [ ] Execute data migration script in Preview environment

satococoa:
Issue identified with migration execution:

1. Current situation:
   - Running `bunx drizzle-kit migrate` completed successfully
   - However, the "type" column was not added to teams table
   - Migration number (0011) appears to conflict with PR #144 changes
   - PR #144's changes (users.email, users.display_name) were already applied

I'm currently evaluating the best approach to resolve this. Will update with the action plan shortly.

satococoa:
I'll hold off on recreating my migration until PR #144 (which adds `users.email` and `users.displayName`) is fully merged into main.

Current plan:
1. Wait for PR #144 migration to be merged
2. Then recreate migration for teams.type column with a new sequence number

This should prevent any migration number conflicts.


cc/ @gentamura

satococoa:
Thank you for merging PR #148.
Now I can proceed with creating a valid migration file.

I'll do these.

- [x] Apply schema migration to Preview environment
- [x] Execute data migration script in Preview environment

```
vi .env.local # add SUPABASE_SERVICE_ROLE_KEY
bun scripts/20241126-db-patch-teams-is-internal-team.ts 
```

```sql
select count(*) from teams where type='internal' -- 130
select count(*) from teams where type='customer' -- 57
select count(*) from teams -- 187
```

satococoa:
## Testing (on Preview)

- [x] signup with @route06.co.jp -> create internal team
- [x] signup with @giselles.ai -> create customer team

satococoa:
@shige 

I'll merge this now.

Could you execute migration and data-migration script?

```
bunx drizzle-kit migrate
vi .env.local # add SUPABASE_SERVICE_ROLE_KEY
bun scripts/20241126-db-patch-teams-is-internal-team.ts 
```

SUPABASE_SERVICE_ROLE_KEY is on the supabase dashboard.
You can find this on Projectsettings > API.
スクリーンショット 2024-12-02 13 36 11

shige:
OK, I will do it from now on!

shige:
Done! @satococoa

PR #126, opened 2024-11-27
Prevent redirection of non-subscribed users when free plan flag is enabled

satococoa:
## Summary
This PR enables users to use Giselle without a subscription.

- Limitations of the free plan will be implemented in other pull requests.
- Normal users won't be affected by this PR because of the feature flag.

PR #133, opened 2024-11-28
Revert "Prevent redirection of non-subscribed users when free plan flag is enabled"

satococoa:
Reverts giselles-ai/giselle#126

There is a bug that internal users can't use giselle without `free-plan` feature flag.


shige (approved):
👍

toyamarinyon (approved):

PR #138, opened 2024-11-28
hotfix: fallback subscription id

toyamarinyon:
better than now

PR #120, opened 2024-11-25
Remove organizations

satococoa:
## Summary
In our new team structure, Organizations is no longer needed.

So I've removed `organizations` table and related codes.

This Pull request is based on https://github.com/giselles-ai/giselle/pull/119 branch.
We must release https://github.com/giselles-ai/giselle/pull/119 first.

## Related Issue
- https://github.com/giselles-ai/giselle/pull/119

## TODO
- [ ] migrate preview
- [ ] migrate production

PR #139, opened 2024-11-28
fix getting team id as measurement scope

satococoa:
## Summary

Originally, this function returns organizationId.
Organization has been removed now, so we use teamId here.

This bug has been introduced here.
- https://github.com/giselles-ai/giselle/pull/120
- https://github.com/giselles-ai/giselle/commit/bd169ff235002f10d8f31725dc9b8b8012d0527a

## Related
- https://github.com/giselles-ai/giselle/pull/120
- https://github.com/giselles-ai/giselle/pull/138


app/(auth)/lib/get-current-subscription.ts
```diff
@@ -1,29 +1,22 @@
...
+export const getUserTeamId = async () => {
 	const user = await getUser();
 	// TODO: When team plans are released, a user may belong to multiple teams, so we need to handle that case.
 	// e.g., fetch team id through agents or so.
-	const [subscription] = await db
-		.select({ id: subscriptions.dbId })
-		.from(subscriptions)
-		.innerJoin(teams, eq(teams.dbId, subscriptions.teamDbId))
+	const [team] = await db
+		.select({ teamId: teams.dbId })
```
shige: 📝 Just My Notes:

I have confirmed the association between `teams.dbId` and `subscriptions` in the following sections.

https://github.com/giselles-ai/giselle/blob/b1c0be70ca2375788859ad39b64d61ed18db7392/drizzle/schema.ts#L54-L56

app/(auth)/lib/index.ts
```diff
@@ -1,7 +1,7 @@
 // export { getUserInitializationTask } from "./get-initialization-task";
 export { deleteOauthCredential } from "./delete-oauth-credential";
 export { getAuthCallbackUrl } from "./get-auth-callback-url";
-export { getUserSubscriptionId } from "./get-current-subscription";
+export { getUserTeamId } from "./get-current-subscription";
```
shige: nits:

I noticed that the file name `get-current-subscription` is outdated, so I was thinking it might be better to update it as well!

shige (approved):
LGTM! 👍

app/(auth)/lib/index.ts
satococoa: @shige 
Thanks!
I renamed it, and found an almost identical function: `getCurrentTeam`.
So, I updated the code to use `getCurrentTeam`.

shige (approved):
LGTM 👍

app/(playground)/p/[agentId]/beta-proto/artifact/server-actions.ts
```diff
@@ -1,10 +1,9 @@
 "use server";
 
-import { openai } from "@ai-sdk/openai";
 import { streamObject } from "ai";
 import { createStreamableValue } from "ai/rsc";
 
-import { getUserSubscriptionId, isRoute06User } from "@/app/(auth)/lib";
+import { getCurrentTeam, isRoute06User } from "@/app/(auth)/lib";
```
Comment (rindrics) [inline app/(playground)/p/[agentId]/beta-proto/artifact/server-actions.ts]: Thanks!

I suggest wrapping `getCurrentTeam()` by `getCurrentSubscriptionUnit()` and import the latter here.

Comment (rindrics) [reply app/(playground)/p/[agentId]/beta-proto/artifact/server-actions.ts]: I think the design below will help future changes in our implementations:
- wrap `getCurrentTeam()` by `getCurrentSubscriptionUnit()`
- place `getCurrentSuscriptionUnit()` to `getCurrentSubscriptionUnit.ts`

Comment (rindrics) [reply app/(playground)/p/[agentId]/beta-proto/artifact/server-actions.ts]: I suggest this because the background motivation of the instrumentation (we want to know usage of external service measured by "subscription unit”) is more stable compared to the design of each subscription plan.

The decision what entity should represent the subscription unit could change depending on our business design; it was originally planned to be an "organization" unit, but has now been changed to a "team".

I believe the change was a good one because the decision was intended to improve the user experience, and we should keep adjusting everything to improve user experience.

Rindrics (commented):
thank you! I placed some comment 👍

app/(playground)/p/[agentId]/beta-proto/artifact/server-actions.ts
Comment (satococoa) [reply app/(playground)/p/[agentId]/beta-proto/artifact/server-actions.ts]: @Rindrics 
Thank you for your suggestion!!

I agree with the abstraction, but I think `subscriptionUnit` is a bit misleading.
"`subscriptionUnit`" strongly suggests fetching records from a subscription table.

This naming implies a direct relationship with subscription data/records which could be misleading, especially since we're also handling free teams that don't have any subscription records.

I'd like to consider another name - do you have any good naming suggestions?

e.g.,

- usageGroup
- resourceScope
- serviceGroup
- teamUnit <- might not be fulfilling its role as an abstraction 🤔 
- unitContext
- resourceBoundary
- usageScope
- measurementScope
- ...

Comment (rindrics) [reply app/(playground)/p/[agentId]/beta-proto/artifact/server-actions.ts]: > "subscriptionUnit" strongly suggests fetching records from a subscription table

yes, I agree.

Thank you for your suggestion!
`measurementScope` seems good to me. @satococoa

Comment (satococoa) [reply app/(playground)/p/[agentId]/beta-proto/artifact/server-actions.ts]: @Rindrics 
Thank you!!
I'll use it.

Comment (satococoa) [reply app/(playground)/p/[agentId]/beta-proto/artifact/server-actions.ts]: done 🚀

Rindrics (approved):
Thank you!

rindrics:
and would you please update the PR title 👍 @satococoa

toyamarinyon (approved):

satococoa:
Verified on Production ✅ 

- Agent is working normally.

PR #140, opened 2024-11-28
Measure usage of external services in web-search node

Rindrics:


rindrics:
I'll follow the changes around `subscriptionId` made by https://github.com/giselles-ai/giselle/pull/139

lib/opentelemetry/wrapper.ts
```diff
@@ -0,0 +1,44 @@
...
+	} catch (error) {
+		if (error instanceof Error) {
+			logger.error(error, `[${externalServiceName}] operation failed`);
+		} else {
+			logger.error(
+				new Error("Unknown error occurred"),
+				`[${externalServiceName}] operation failed`,
+			);
+		}
+		throw error;
```
shige, line 44: @Rindrics 

In the case of errors occurring during OpenTelemetry instrumentation, wouldn't it be better not to throw an error?  
In my opinion, errors related to observability should not interrupt business logic.  
Instead of throwing an error, I believe it's sufficient to send the error to a bug tracking system like Sentry.

toyamarinyon (approved):

lib/opentelemetry/wrapper.ts
rindrics: yes, you're right!

Instrumentation MUST NOT interfere our business logic.

I'll correct the error case.

lib/opentelemetry/wrapper.ts
```diff
@@ -1,44 +1,46 @@
...
+						externalServiceName,
+						requestCount: 1,
+						duration,
+						measurementScope,
+						isR06User,
+					};
+					logger.info(metrics, `[${externalServiceName}] response obtained`);
+				})
+				.catch((getMetricsTagError) => {
+					captureError(
```
rindrics, line 31: won't raise error not to interfere business logic

lib/opentelemetry/wrapper.ts
```diff
@@ -1,44 +1,46 @@
...
+						logger,
+						getMetricsTagError,
+						"failed to get user info for logging",
+					);
+				});
+		} catch (instrumentationError) {
+			captureError(logger, instrumentationError, "instrumentation failed");
```
rindrics, line 38: won't raise error not to interfere business logic

lib/opentelemetry/wrapper.ts
rindrics: I fixed the logging wrapper by https://github.com/giselles-ai/giselle/pull/140/commits/54379563a4c8a8620944d7bc9bb03eb6169aa2a6, but remained this final error throwing as it was because this error is thrown by the business-logic function (passed as `operation` to the logging wrapper). @shige 

On the other hand, I corrected other instrumentation-related errors not to interfere business logic and to be just sent to Sentry

shige (approved):
LGTM 👍 Thanks!

rindrics:
thx!

PR #162, opened 2024-12-04
fix(shortcuts): Prevent keyboard shortcuts in form inputs

toyamarinyon:
Add check to ignore keyboard shortcuts when focus is on input elements:
- Add list of ignored HTML tags (INPUT, TEXTAREA, SELECT)
- Check active element before triggering shortcuts
- Prevents unintended tool switching while typing

This fixes potential conflicts between keyboard shortcuts and text input, improving usability when editing text fields or forms.


shige (approved):
LGTM! 🆙

PR #166, opened 2024-12-04
feat(ui): Add generate text functionality to properties panel

toyamarinyon:
Add text generation controls and empty state UI with two main actions:
- Generate text using current prompt
- Navigate to prompt editor

The update improves UX by providing clear actions when no text is generated and streamlines the text generation workflow.

> [!NOTE]
> The lengthy process of generate has increased to two locations, which will be summarised in a separate Pull request.


shige (approved):
LGTM 👍

PR #182, opened 2024-12-06
feat(playground): Add customizable system prompt for text generation

toyamarinyon:
https://github.com/user-attachments/assets/a4932471-2ee5-4da8-9fad-47b0990dae01

> [!NOTE]
> The fact that the text is not changed when “Revert to Return” is pressed is an unnatural behavior and should be corrected separately.

Add ability to customize and reset system prompts in text generation nodes:
- Add system prompt field to TextGenerateActionContent type
- Create collapsible system prompt editor with persistence
- Add revert to default functionality with visual feedback
- Use default prompt when system is undefined

This gives users more control over LLM behavior while maintaining the ability to return to default settings.

### DIFF
--- a/app/(playground)/p/[agentId]/canary/actions.ts
+++ b/app/(playground)/p/[agentId]/canary/actions.ts
@@ -243,7 +243,9 @@ export async function action(
 			const actionSources = await resolveSources(node.content.sources);
 			const requirement = resolveRequirement(node.content.requirement);
 			const model = resolveLanguageModel(node.content.llm);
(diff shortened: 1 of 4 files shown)


shige (approved):
LGTM! 👍

toyamarinyon:
I forgot to reflect the settings in the UI in the Action, so I did it in d8d10ae931d331b7fbfdfe5a44657ad688804a4a.

shige (approved):
LGTM! 👍

PR #193, opened 2024-12-09
[feat] Implement graph versioning and subgraph organization system

toyamarinyon:
It is prepared to run a graph composed of nodes and their connections that are placed in Playground in a collective manner.

## Changes
- Add versioning system for graph schemas with "2024-12-09" format
- Implement automatic migration system for legacy graphs

    Migration is performed automatically when user open the Playground with legacy graphs. No migration scripts needed!

- Create subgraph derivation logic using depth-first search
- Add developer tools tab with graph download capability

    **Visible Only Developer Mode Flag On**

    image

- Introduce TabValueContext for navigation state management
- Reorganize graph utility functions into separate modules


shige (approved):
LGTM! ⚒️

Issue #39, opened 2024-10-23
Datetime in the database is off by 9 hours


satococoa:
### Self Checks

- [x] This is only for bug report, if you would like to ask a question, please head to [Discussions](../discussions/categories/q-a).
- [x] I have searched for existing issues [search for existing issues](../issues), including closed ones.

### Cloud or Self Hosted

Cloud

### Steps to reproduce

1. Signup
2. Purchase a Subscription
3. The `created` timestamp differs between Stripe and the database

On Stripe:
![Image](https://github.com/user-attachments/assets/e093475b-8d15-4b90-8406-c0ddb6d5539f)

### Expected Behavior

The `created` value in the database should be `2024-10-23T00:07:12.000Z`.

### Actual Behavior

In the database:
![Image](https://github.com/user-attachments/assets/ac64d382-2633-49d0-84f2-d8b2fa9afb44)

satococoa:
# Timestamp display issue investigation report

## Summary
The timestamps stored in the database are **correct**, while the timestamps displayed in Drizzle Studio and the Vercel console are incorrect.

## Verification process
1. Database Validation
   * Compared timestamps between our database and Stripe's API data
   * Used a script to verify all subscription records
   * Found perfect match between database records and Stripe data
2. Raw Timestamp Verification
   * Extracted raw epoch timestamps from the database
   * Example query:
   ```sql
   SELECT 
     id,
     extract(epoch from created) as created_epoch,
     extract(epoch from ended_at) as ended_epoch
   FROM subscriptions 
   WHERE id='sub_1QCs6W2MBZMnjD8ta61FpLuv';
   ```
   * Confirmed that epoch timestamps exactly match Stripe's data

## Testing script

<details><summary>verify script</summary>
<p>

```ts
import { db, subscriptions } from "@/drizzle";
import { stripe } from "@/services/external/stripe";
import type Stripe from "stripe";

async function main() {
	const stripeSubscriptions = await fetchStripeSubscriptions();
	console.log("stripeSubscriptions", stripeSubscriptions.length);

	const dbSubscriptions = await db.select().from(subscriptions);
	console.log("dbSubscriptions", dbSubscriptions.length);

	for (const stripeSubscription of stripeSubscriptions) {
		const dbSubscription = dbSubscriptions.find(
			(it) => it.id === stripeSubscription.id,
		);
		if (!dbSubscription) {
			continue;
		}

		if (
			dbSubscription.created.getTime() !==
[...]

PR #200, opened 2024-12-10
Add Agent Time Usage Aggregation

satococoa:
## Summary
This PR implements functionality to aggregate and display agent execution time per team.

スクリーンショット 2024-12-10 15 20 00

## Changes
- Implemented agent execution time aggregation logic
  - Get total execution time by team ID and time period
  - Store execution time in milliseconds, display in minutes
- Handle period aggregation as half-open interval  [startedAt, endedAt)
  - Includes start time, excludes end time
- Add indexes: `agents.teamDbId`, `agentActivities.endedAt`

## TODO

 - [x] migration for Preview
 - [ ] migration for Production


app/(main)/settings/team/agent-time-charge.tsx
```diff
@@ -0,0 +1,23 @@
+import { calculateAgentTimeUsage } from "@/services/agents/activities/agent-time-usage";
+import { fetchCurrentTeam } from "@/services/teams";
+import {
+	AgentTimeUsageForFreePlan,
+	AgentTimeUsageForProPlan,
+} from "./components/agent-time-usage";
+
+export async function AgentTimeCharge() {
+	const currentTeam = await fetchCurrentTeam();
+	const isProPlan =
```
shige: I thought it would be better to use the following!

https://github.com/giselles-ai/giselle/blob/53ed3f50aa76f4a1accb5d897687e34cb037dcfe/services/teams/utils.ts#L3-L5

app/(main)/settings/team/components/agent-time-usage.tsx
```diff
@@ -0,0 +1,158 @@
...
+	/** Message displayed when approaching threshold */
+	warningMessage?: string;
+	/** Message displayed when exceeding limit */
+	overLimitMessage?: string;
+};
+
+function AgentTimeUsage(props: AgentTimeUsageProps) {
+	const {
+		usedMinutes,
+		includedMinutes = 120,
```
shige: The value `120` appears multiple times, so I think it would be better to define it as a constant, at least within this file.

services/agents/activities/utils.ts
```diff
@@ -0,0 +1,81 @@
+export function getMonthlyBillingCycle(
```
shige, line 1: Question:

Is this method guaranteed to work regardless of the timezone in which it is executed?

services/agents/activities/agent-time-usage.ts
```diff
@@ -0,0 +1,102 @@
...
+ * @returns Promise that resolves to the total agent time usage in minutes
+ * @example
+ * const timeUsage = await calculateAgentTimeUsage(123);
+ * console.log(timeUsage); // 42.8
+ */
+export async function calculateAgentTimeUsage(teamDbId: number) {
+	const { start, end } = await getCurrentBillingPeriod(teamDbId);
+	const timeUsageMs = await agentTimeUsageMs(teamDbId, start, end);
+
+	return Math.ceil((timeUsageMs / 1000 / 60) * 10) / 10;
```
shige: IMO:

At first glance, since this involves time calculations that affect pricing, I thought it might be safer to use a library like the ones below.  
On the other hand, since it’s not directly calculating the amount, this approach might be fine as it is.

- [big\.js vs bignumber\.js vs decimal\.js vs mathjs vs numeraljs \| npm trends](https://npmtrends.com/big.js-vs-bignumber.js-vs-decimal.js-vs-mathjs-vs-numeraljs)

shige:
@satococoa 
The review is still in progress, but since this is a PR involving time calculations that affect pricing, I anticipate it will take longer than usual to review. For this reason, I’ve been commenting on it step by step.  

In relation to this, does the following issue need to be addressed? Will it affect the time calculations?  

- https://github.com/giselles-ai/giselle/issues/39

services/agents/activities/agent-time-usage.ts
satococoa: I plan to use this method to calculate the pricing affecting value.
Therefore, I'll move the rounding logic closer to the UI layer.

services/agents/activities/utils.ts
satococoa: Thank you. The answer is "No".

- Given `referenceDate` is from the database which is stored in UTC. So this may be okay.
- Given `curerntDate`, I passed `new Date()` and this is wrong. This must be fixed.

I think using a good library like `date-fns` or calculating all values as UTC is better to ensure accurate value.
I'll change this.

----

And timestamp value in the database must be fixed.
ref https://github.com/giselles-ai/giselle/issues/39

satococoa:
@shige 

> In relation to this, does the following issue need to be addressed? Will it affect the time calculations?

Yes. The timestamps stored in the database is wrong now. This must be fixed for our agent time charge calculation.

services/agents/activities/utils.ts
satococoa: > Given referenceDate is from the database which is stored in UTC. So this may be okay.

Sorry, ⬆️ This my understanding was wrong, also.

```ts
// in JST
const d = new Date("2024-07-31T19:17:10.156Z")
d.getDate() // => 1
d.getUTCDate() // => 31
```

I thought `d.getDate()` should return `31` but it does not.
We must use the `getUTCXXX` version.

app/(main)/settings/team/agent-time-charge.tsx
satococoa: ⚒️  [adcf400](https://github.com/giselles-ai/giselle/pull/200/commits/adcf400da63095c9cb97c8b283a36ef1abd6c15a)

app/(main)/settings/team/components/agent-time-usage.tsx
satococoa: 🛠️ https://github.com/giselles-ai/giselle/pull/200/commits/6fc7b1bd26dff1ec8ff378fdb4a39b5b163b2e54

services/agents/activities/agent-time-usage.ts
satococoa: 🛠️ [1a6d4d2](https://github.com/giselles-ai/giselle/pull/200/commits/1a6d4d2074fb0819d5ffd1f4c6c3181cc9238fc2)

package.json
```diff
@@ -65,6 +65,7 @@
 		"clsx": "2.1.1",
 		"cmdk": "^1.0.0",
 		"cva": "1.0.0-beta.1",
+		"decimal.js": "^10.4.3",
```
shige: Question @satococoa 

Was it difficult to achieve this with `big.js`?  
From the following points of view, I think it might be better to use `big.js` if there are no issues:

- Actively maintained  
- Smaller file size  

https://npmtrends.com/big.js-vs-bignumber.js-vs-decimal.js-vs-mathjs  

image

satococoa: @shige 

I chose decimal.js because it includes type definitions, whereas big.js doesn't (it needs `@types/big.js`).
However, big.js (with @types/big.js) would also work if the type definitions are good enough.

shige: @satococoa 
I understand the background!  
For now, I think it's fine as it is. 👌

toyamarinyon (commented):
Why do we need decimal.js? Can I implement and test the required implementation using LLM?

shige:
@toyamarinyon 

The reason is that I do not want to perform floating-point calculations related to fee calculation in TypeScript.

> Can I implement and test the required implementation using an LLM?

I think it might be possible if I add the source code or documentation of `decimal.js` as context, but I haven’t tried it yet.

toyamarinyon:
@shige I see. So what about calculating in SQL?

shige:
@toyamarinyon 

Sorry, I didn't understand the intent of your question.  
Are you asking about performing the cost calculation logic required for expense calculation in SQL?  

If that's the case, it could end up growing into something like a stored procedure, and honestly, I don't even want to think about it.

services/agents/activities/utils.ts
satococoa: I've fixed timezone issue with additional test cases.

satococoa:
@toyamarinyon @shige 

In this PR, the numbers are only displayed to the user, so strict precision is not required.
Therefore, it may not be necessary to add decimal.js at this point.

However, in the near future, when we implement billing calculations, I would prefer to write the rounding logic and other calculation logic in TypeScript rather than SQL. This is for the sake of easier testing and centralized management of business logic.

Therefore, I would like to propose the following:

- Do not add decimal.js at this point
- Reconsider the approach when writing the precise billing calculation logic
  - SQL (aggregation functions)
  - decimal.js
  - big.js or other libraries

shige:
@satococoa 
Sorry, I wrote various things, but I now think this approach is fine. 🙏 

> * Do not add decimal.js at this point
> * Reconsider the approach when writing the precise billing calculation logic
>   
>   * SQL (aggregation functions)
>   * decimal.js
>   * big.js or other libraries

services/agents/activities/utils.test.ts
toyamarinyon, line 1: The test cases added to `utils.test.ts` cover various scenarios for the `getMonthlyBillingCycle` function, including different time zones, daylight saving time transitions, month-end cases, and environment independence.

However, there are a few potential omissions:
1. **Leap Year Handling:** Test cases for dates during a leap year (e.g., handling February 29).
2. **Invalid Date Handling:** Test cases to handle invalid date inputs or edge cases where the input dates are not valid.
3. **Boundary Conditions:** Additional cases testing boundary conditions, like the very last second of a month.

shige (approved):
LGTM 👍 

I will do my OATs after migrating to the preview environment.

services/agents/activities/utils.test.ts
toyamarinyon: You don't have to, but you may want to add it if you can quickly.

satococoa: Thank you.
I've added [225d555](https://github.com/giselles-ai/giselle/pull/200/commits/225d5555cc9bb563b3de7dfa4d385ace2f7434ed).

satococoa:
@shige I've executed migration for Preview.

shige:
> I've executed migration for Preview.

I checked it. Thank you!

toyamarinyon (approved):

satococoa:
Thank you for your review 🚀 

@shige 
Could you run the migration for Production and deploy this?

shige:
@satococoa 

I did it! `bun drizzle-kit migrate` 🚀

PR #204, opened 2024-12-11
feat(llm): Add Google Gemini model support

toyamarinyon:
Add Google AI SDK integration and Gemini 1.5 model options to the language model selection interface. This enables users to leverage both Gemini 1.5 Flash and Pro models alongside existing OpenAI and Anthropic options.

- Add @ai-sdk/google dependency
- Implement Google provider resolution in actions.ts
- Update model selector UI with Gemini options


shige (approved):
LGTM! ♊

PR #207, opened 2024-12-11
add feedback block

koji-m:


shige:
We will implement this separately, so we will close this issue.  
Thank you very much!

PR #209, opened 2024-12-11
Add agent time usage limit for free plan teams

satococoa:
## Summary
Added limitation that prevents free plan teams from running workflows when their agent time charge has been depleted.

## Changes
- Added `hasEnoughAgentTimeCharge()` function to check if a team has enough agent time remaining
- If the limit is exceeded, `executeFlow()` will throw an error with a user-friendly message
  - Instead of adding a simple alert it would be better to use toast notifications or error states. However, since v2 playground development is ongoing, it makes more sense to implement improved error notifications in v2.

## Screenshots
スクリーンショット 2024-12-11 14 50 48


toyamarinyon (approved):

shige (approved):
LGTM! ⏲️

Issue #208, opened 2024-12-11
Reduce the number of calls to `supabase.auth.getUser()` per request.


shige:
## Task description
Reduce the number of calls to `supabase.auth.getUser()` per request.

## Background
Currently, the application makes multiple calls to `supabase.auth.getUser()` for each request, which can lead to performance issues and increased latency. By minimizing the number of calls, we can enhance the application's efficiency and improve the user experience. This task is necessary to streamline authentication handling and reduce unnecessary load on the Supabase service.

## TODO (Optional)
- [ ] Analyze the current implementation to identify how many times `supabase.auth.getUser()` is called per request.
- [ ] Refactor the code to consolidate calls to `supabase.auth.getUser()` where possible.
- [ ] Test the changes to ensure that user authentication information is still retrieved correctly.
- [ ] Document the changes made and any improvements in performance.

## Additional notes (Optional)

satococoa:
It seems to be better now.
If you notice something weird, please feel free to reopen or create new one.

PR #210, opened 2024-12-11
Optimize heavily used functions with React.cache

satococoa:
ref #208

## Summary
Optimize function calls using [React's cache](https://react.dev/reference/react/cache).

Since caches created by React's cache are only valid within the same request, this optimization is safe and won't cause any stale data issues.

## Implementation Details
- Wrapped `getUser`,  `fetchCurrentUser`, `fetchCurrentTeam` functions with `cache` to memoize results
- Cache is automatically invalidated between different requests
- Original error handling and retry logic remains unchanged

## Alternative Approach Considered
I initially explored overriding Supabase's fetch implementation with Next.js's fetch.
However, there is a known issue that requires careful consideration:
https://github.com/supabase/supabase-js/issues/917

Therefore, I opted to use React's cache as it provides a more straightforward solution.

## Testing
- [x] Verified cache behavior in server components
- [x] Confirmed error handling works as expected
- [x] Tested with different user IDs to ensure proper caching

### DIFF
--- a/lib/supabase/get-user.ts
+++ b/lib/supabase/get-user.ts
@@ -1,4 +1,5 @@
 import { isAuthRetryableFetchError } from "@supabase/supabase-js";
+import { cache } from "react";
 import { withRetry } from "../utils";
 import { createClient } from "./server";
 
@@ -11,7 +12,7 @@ import { createClient } from "./server";
  * IMPORTANT: This function will throw an error if executed while the user is not authenticated.
  * Make sure the user is logged in before calling this function.
  */
-export const getUser = async () => {
+const getUser = async () => {
 	const supabase = await createClient();
 
 	const getUserFunc = async () => {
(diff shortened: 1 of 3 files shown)


shige (approved):
LGTM! ⚡ 

Thanks for improving so quickly!

toyamarinyon (commented):
I can't decide at a quick glance whether this is a case where I should use React.cache, and I'm thinking of taking the time to review it, but should I just merge it and see what happens?

shige:
@toyamarinyon 

Currently, even just making a request to the `/settings/team` endpoint results in  
https://github.com/giselles-ai/giselle/blob/e93f5113841cdbdecab2bc3de4ece360518bfc9b/lib/supabase/get-user.ts#L18
being called 12 times.

Therefore, in my opinion, I’d like to release it and monitor the situation.  
If there are any other improvement ideas, I’d be happy if you could work on them! 🧑‍💻

PR #213, opened 2024-12-11
[refactor] Simplify action execution by removing explicit graph URL parameter

toyamarinyon:
## Changes
- Remove `graphUrl` parameter from execute action signature
- Update execution context to no longer manage graph URLs
- Simplify execution flow in execution provider component
- Adjust server action to handle graph URL fetch internally

## Why
This change streamlines the execution flow by removing the need to pass graph URLs through the client. The graph URL is now managed entirely server-side through database lookups using the agent ID, which:
- Improves security by preventing client-side URL manipulation
- Reduces complexity in the execution context
- Ensures single source of truth for graph URLs in the database

### DIFF
--- a/app/(playground)/p/[agentId]/canary/actions.ts
+++ b/app/(playground)/p/[agentId]/canary/actions.ts
@@ -1,5 +1,6 @@
 "use server";
 
+import { db } from "@/drizzle";
 import { anthropic } from "@ai-sdk/anthropic";
 import { google } from "@ai-sdk/google";
 import { openai } from "@ai-sdk/openai";
@@ -16,6 +17,7 @@ import * as v from "valibot";
 import { vercelBlobFileFolder, vercelBlobGraphFolder } from "./constants";
 import { textGenerationPrompt } from "./prompts";
 import type {
+	AgentId,
 	ArtifactId,
 	FileData,
 	FileId,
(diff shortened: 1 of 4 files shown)


app/(playground)/p/[agentId]/canary/actions.ts
```diff
@@ -106,15 +108,22 @@ type ActionSource = TextSource | TextGenerationSource | FileSource;
...
 	const lf = new Langfuse();
 	const trace = lf.trace({
 		sessionId: artifactId,
 	});
 
-	const graph = await fetch(graphUrl).then(
+	const agent = await db.query.agents.findFirst({
+		where: (agents, { eq }) => eq(agents.id, agentId),
+	});
+	if (agent === undefined || agent.graphUrl === null) {
```
Comment (shige) [inline app/(playground)/p/[agentId]/canary/actions.ts:122]: nits:

Wouldn't it be possible to specify `agent.graphUrl` in the query at line 120 instead of determining it here?

shige (approved):
LGTM 👍 

From the perspective of refactoring, there was one point that caught my attention, so I left a comment about it.

app/(playground)/p/[agentId]/canary/actions.ts
Comment (toyamarinyon) [reply app/(playground)/p/[agentId]/canary/actions.ts:122]: As it is a Nullable column, the type definition will be Null even if it is narrowed down in a query.

This will be changed in the future as it will no longer be needed if it is set to Not null.

https://github.com/giselles-ai/giselle/blob/cce94aa476c30265fc62fdf5d3ae1f38a2d308e9/drizzle/schema.ts#L135-L136

Issue #217, opened 2024-12-12
Apply instrumentations for usage of external services to 'canary' version


Rindrics:
## Task description

Apply preexistent instrumentations to 'canary' version to measure usage of external services

## Background

- To optimize plan fee and user experience, we added instrumentations to measure usage of external services by issues below:
  - #128 
  - #141 
  - #197 
  - #198
- The instrumentation was only works on current "beta-proto "version
- We have to apply it for internal newer version "canary" to prepare for its release

## TODO (Optional)

Apply preexistent instrumentations to the use cases below:
- [x] #218
- [x] #219
- [x] #220

shige:
We're in the midst of a re-architecture, so we won’t be addressing it and will close as not planned.

- ref: https://github.com/giselles-ai/giselle/discussions/284

PR #221, opened 2024-12-12
Measure usage of text-generation feature of external services on "canary" agent

### LINKED ISSUE #218: [canary] Measure usage of external services that provides text generation
we use  text generation feature of external services inside:
- text-generation node (core feature)
- web-search node (to create search keywords)

Rindrics:
## Summary

This PR enables usage of external services that provides text-generation feature in "canary" version agent

## Related Issue

- This PR closes #218
  - as subtask of #217

## Changes

- apply preexistent instrumentation to "canary" agent
- update instrumentation to get names of  external services and models dynamically

## Testing

Using corresponding preview (private) environment, arrival of telemetry data was confirmed on the following use case

#### text-genertion ✅ 

image
image

#### web-search ✅ 

image
image

#### text-generation with OpenAI ✅ 
image

#### text-generation with Claude ✅ 

image

#### web-search with  Gemini ✅ 

image


lib/opentelemetry/wrapper.ts
```diff
@@ -1,15 +1,57 @@
...
+				externalServiceName: ExternalServiceName.Google,
+				modelId: model.modelId,
+			};
+		default:
+			logger.error(
+				new Error(`unknown model class name '${modelClassName}' obtained`),
+				"consider adding to 'ExternalServiceName'",
+			);
+			return {
+				externalServiceName: UnimplementedServiceName.TODO,
```
rindrics: `UnimplementedServiceName.TODO` is defined to fulfill type safety, not for actual runtime use case.

This workaround is required because we don't want to interfere business logic with raising error.

lib/opentelemetry/types.ts
```diff
@@ -2,24 +2,46 @@ import { Strategy } from "unstructured-client/sdk/models/shared";
...
 	OpenAI: "openai",
 	Tavily: "tavily",
 	Unstructured: "unstructured",
 	VercelBlob: "vercel_blob",
 } as const;
 
-export type ExternalServiceName = // Name of the service to which agent requests
-	(typeof ExternalServiceName)[keyof typeof ExternalServiceName];
+export const UnimplementedServiceName = {
+	TODO: "todo", // for type safety
```
shige:
```suggestion
	Unknown: "unknown", // for type safety
```

shige: nits: 

The text reads "TODO," but I don't think there's actually anything to be done here. I think something like "Unimplemented" or "Unknown" would be more appropriate.  

Incidentally, "Unknown" was a word suggested by Copilot.

https://github.com/giselles-ai/giselle/pull/221/files#diff-49a97b0518ec7e083f544888dcd1be31d3b81c5f68775c2c465041ae9def9fdbR58 is also "unknown".

shige (commented):
@Rindrics 

> ### "canary" agent
> #### text-generation with

It seems that OpenAI is not supported at the moment, but is it planned to be implemented in another pull request?

lib/opentelemetry/types.ts
rindrics: Thanks! The suggestion applied: https://github.com/giselles-ai/giselle/pull/221/commits/7e6a7cf1e35f963b121531b101ab3ebf6487a224

rindrics:
> It seems that OpenAI is not supported at the moment, but is it planned to be implemented in another pull request?

Sorry, it does support! @shige

I cut corner of showing the "OpenAI" case in canary agent because the mechanism of getting provider name is common between agent types.

I added the "OpenAI" case to the description 👍

shige (approved):
LGTM! 👍

app/(playground)/p/[agentId]/canary/actions.ts
```diff
@@ -80,6 +87,14 @@ const artifactSchema = v.object({
...
 		),
 	),
+	usage: v.pipe(
+		v.object({
+			promptTokens: v.number(),
+			completionTokens: v.number(),
+			totalTokens: v.number(),
+		}),
+		v.description("Return from 'generateObject()' from @vercel/ai"),
+	),
```
Comment (toyamarinyon) [inline app/(playground)/p/[agentId]/canary/actions.ts]: This will cause the LLM to generate the USAGE, so you will not get the USAGE you want to get; I think you should use the return USAGE of the streamObject, so I created a reference implementation below.

https://github.com/giselles-ai/giselle/pull/229

Comment (rindrics) [reply app/(playground)/p/[agentId]/canary/actions.ts]: thanks, that looked good!

Let us merge that into here when build error is resolved.

toyamarinyon (approved):

rindrics:
thx!

PR #224, opened 2024-12-12
fix: Simplify team member list view and state management

gentamura:
## Summary
Improve team member list component by simplifying state management and component re-rendering logic. This change makes the component behavior more predictable and maintainable.

## Related Issue
N/A

## Changes
- Replace currentUserRoleState tuple with direct currentUserRole prop
- Remove isUpdatingSelf from action responses
- Move role update logic to parent component for better state management
- Use key prop with multiple variables for fine-grained control of component re-rendering

## Testing
- Verified team member list updates correctly when switching teams
- Confirmed role updates work properly
- Checked component re-rendering behavior with key changes
- Tested member deletion functionality

## Other Information
This refactoring aligns with React's best practices by:
- Simplifying state management
- Making component updates more predictable
- Improving code maintainability


app/(main)/settings/team/actions.ts
```diff
@@ -269,7 +255,6 @@ export async function updateTeamMemberRole(formData: FormData) {
 
 		return {
 			success: true,
-			isUpdatingSelf,
```
gentamura, line 272: `isUpdatingSelf` was removed because it was originally introduced for optimistic updates, but was not used as expected in the previous implementation, so it was removed to clean up the state management.

app/(main)/settings/team/team-members-list-item.tsx
```diff
@@ -29,23 +29,22 @@ type TeamMemberListItemProps = {
 	displayName: string | null;
 	email: string | null;
 	role: TeamRole;
-	currentUserRoleState: [TeamRole, (role: TeamRole) => void];
```
gentamura, line 32: The parent component was introduced with the assumption that `currentUserRole` would be updated, but since re-rendering is now managed by key, this is no longer necessary, so `currentUserRole` is now used directly.

app/(main)/settings/team/team-members-list.tsx
```diff
@@ -30,12 +25,12 @@ export function TeamMembersList({
 			<div className="divide-y divide-zinc-800">
 				{members.map((member) => (
 					<TeamMemberListItem
-						key={member.userId}
+						key={`${member.userId}-${member.displayName}-${member.role}-${currentUserRole}`}
```
gentamura: If the same user belongs to multiple teams and the key is only userId, the data is not updated after switching teams.
Fixed key configuration to handle cases where referenced data differs by team affiliation.

ref: https://react.dev/learn/you-might-not-need-an-effect#resetting-all-state-when-a-prop-changes

gentamura:
@satococoa review, please :pray:

app/(main)/settings/team/team-members-list.tsx
satococoa: How about adding `teamDbId` to `TeamMembersListProps` and using it in the key prop?:

``key={`${teamDbId}-${member.userId}`}``

satococoa (commented):
Thank you!
I've left a comment 👀

app/(main)/settings/team/team-members-list.tsx
gentamura: @satococoa Thanks 👍  b3cc5e50aece fixed

satococoa (approved):
Thank you 😄 👍

gentamura:
@shige @toyamarinyon review, please 🙏

docs/packages-license.md
toyamarinyon, line 1: 📝 Cause https://github.com/giselles-ai/giselle/pull/227

toyamarinyon (approved):

app/(main)/settings/team/team-members-list.tsx
```diff
@@ -30,12 +27,12 @@ export function TeamMembersList({
 			<div className="divide-y divide-zinc-800">
 				{members.map((member) => (
 					<TeamMemberListItem
-						key={member.userId}
+						key={`${teamDbId}-${member.userId}`}
```
shige, line 30: IMO:

Using `teamDbId` in client-side code is not ideal, but since it's only being used as a key for the DOM, I think it's fine to leave it as is.

shige (approved):
LGTM! 👍

app/(main)/settings/team/team-members-list.tsx
gentamura: This is about using public id and db id, right? I too thought that the public id was originally needed, but since it is removed by key, I figured it was no problem 👌

gentamura:
Thanks for reviewers 🚀

PR #227, opened 2024-12-13
chore(deps): Upgrade AI SDK packages for compatibility

toyamarinyon:
I updated the AI SDK packages to try streamText transform option:

> Many LLM providers send large response chunks, which can lead to choppy user experiences.
>
> With smoothStream, you can stream out full words with a configurable delay:
> https://x.com/aisdk/status/1867239473375047791

this option is available for only the `streamText` method, we can't use it for `streamObject` method.

but, I want to update ai sdk packages to the latest version and fix stream handling code to maintain compatibility with the updated SDKs.


shige (approved):
LGTM! 👍

PR #232, opened 2024-12-13
feat(playground): Add agent duplication feature with file handling

toyamarinyon:
Add functionality to clone agents with their associated files and graphs:

- Create new server action `copyAgentAction` for agent duplication
- Handle file copying using Vercel Blob storage
- Generate new IDs for duplicated files and graphs
- Add UI form component for agent ID input with validation
- Implement loading and success states for better UX


shige:
### ⚠️ Disclaimer

This feature is intended for use in customer support and debugging purposes.  
Its primary purpose is to be utilized, with user consent, in cases where an Agent cannot execute properly.  
Initially, this feature will be used internally for validation purposes.

shige (approved):
LGTM! 👍

PR #234, opened 2024-12-16
security(headers): Add X-Frame-Options header to prevent clickjacking

toyamarinyon:
Add security header configuration to Next.js to protect against clickjacking attacks by setting X-Frame-Options: DENY for all paths matching /path*. This prevents the application from being embedded in iframes on other domains.

Security: This is a recommended security best practice to mitigate cross-frame scripting attacks.

### DIFF
--- a/next.config.ts
+++ b/next.config.ts
@@ -26,6 +26,19 @@ const nextConfig: NextConfig = {
 			},
 		];
 	},
+	async headers() {
+		return [
+			{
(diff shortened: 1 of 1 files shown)


toyamarinyon:
I've checked it works on preview env

Screenshot 2024-12-16 at 12 42 25

shige (approved):
LGTM! 👍

Issue #130, opened 2024-11-28
Client error occurs sometimes after page transition


Rindrics:
## Task description

- resolve cause of client error (happens rarely) related to authentication

## Background

- sometimes client error occurs after page transition
- the error is due to repetitive failure in fetching user data from external auth backend (we use Supabase)

## Workaround available for users

- users can get page back to normal state by refreshing the page

## Resolution candidates

one listed above is more desirable:
- resolve the root cause of the fetch error (if possible)
- reduce number of fetch attempts to avoid [`MaxRetriesExceedError`](https://github.com/giselles-ai/giselle/blob/59e77d0782dd6eccc1437883d9f350a8ea1e4dd5/lib/utils.ts#L17)
- increase [maxRetry](https://github.com/giselles-ai/giselle/blob/59e77d0782dd6eccc1437883d9f350a8ea1e4dd5/lib/utils.ts#L47) to decrease possibility of `MaxRetriesExceedError` occurrence

## Additional notes (Optional)

related comment: https://github.com/giselles-ai/giselle/pull/107#pullrequestreview-2464838317

shige:
@sakit0 I have assigned this issue to you as a good first issue!

sakit0:
The `cache()` has been added to the data fetch function, so the problem may not occur.

thearnabsaha:
can i work on this?

shige:
This issue is outdated so I'm closing it.

PR #237, opened 2024-12-16
Add caching to `getUser` to reduce repeated fetch requests

sakit0:
## Summary
This PR introduces caching for the `getUser` function. By using React's `cache()` API, it prevents repeated fetch requests for the user data, thereby reducing unnecessary network calls and mitigating recurring `isAuthRetryableFetchError` issues.

## Related Issue
https://github.com/giselles-ai/giselle/issues/130

## Changes
- Wrapped the `getUser` function with `cache()` to memoize results.

## Testing
- manual test

### DIFF
--- a/lib/supabase/get-user.ts
+++ b/lib/supabase/get-user.ts
@@ -13,32 +13,32 @@ import { createClient } from "./server";
  * Make sure the user is logged in before calling this function.
  */
 const getUser = cache(async () => {
-  const supabase = await createClient();
+	const supabase = await createClient();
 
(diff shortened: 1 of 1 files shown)


toyamarinyon (commented):
Could you help me understand the caching mechanism?
I see that caching is already implemented in the original code, but are you suggesting that caching wasn't working in previous Pull requests?

sakit0:
@toyamarinyon 
Sorry, I noticed that the output result is still the same. 
I will change it!

PR #240, opened 2024-12-16
feat(agent): Add unique graph ID generation for agent copies

toyamarinyon:
Modify copyAgentAction to generate new graph IDs when copying agents to ensure each copy has a distinct graph identity. This prevents potential conflicts and maintains data isolation between copied agents.

Changes:
- Import createGraphId utility
- Generate new graph ID during copy operation
- Pass new ID when creating graph copy


shige (approved):
LGTM! 🆔

PR #243, opened 2024-12-17
fix(graph): Handle ghost connections in graph flow derivation

toyamarinyon:
Add protection against invalid node references in connections and update graph version to 20241217

- Skip connections that reference non-existent source/target nodes
- Add test case to verify ghost connection handling
- Update graph version to 20241217 with migration path
- Update initialization and version checking logic

Fixes potential errors when processing broken graph connections


shige:
@toyamarinyon It seems the CI has failed. Could you fix it, please?

toyamarinyon:
Sorry! I'll fix it quickly!

shige (approved):
LGTM! 👍

PR #245, opened 2024-12-17
fix: Resolve type error in agent activity duration calculation

gentamura:
## Summary
Fix a TypeScript build error related to the calculation of agent activity duration. The error occurred because the type of `activity.usedCharge` wasn't properly handled in arithmetic operations.

image

## Related Issue
N/A

## Changes
- Add type safety for `activity.usedCharge` in duration calculations
- Implement explicit number type checking
- Add error handling for invalid duration values
- Return 0 for invalid duration values with error logging

## Testing
- Verified build succeeds with the type error fixed
- Confirmed calculation results are correct for valid inputs

## Other Information
The fix ensures type safety while maintaining the original functionality. We chose to handle invalid values by defaulting to 0 and logging errors rather than throwing exceptions to maintain application stability.

### DIFF
--- a/app/(main)/settings/team/actions.ts
+++ b/app/(main)/settings/team/actions.ts
@@ -450,11 +450,19 @@ export async function getAgentActivities({
 			.orderBy(desc(agentActivities.startedAt))
 			.limit(limit);
 
-		const formattedActivities = activities.map((activity) => ({
-			...activity,
-			// Convert milliseconds to seconds and round to 2 decimal places
-			usedCharge: Math.ceil((activity.usedCharge / 1000) * 100) / 100,
-		}));
+		const formattedActivities = activities.map((activity) => {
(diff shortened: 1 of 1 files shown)


gentamura:
@shige @toyamarinyon @satococoa review, please 🙏

toyamarinyon (approved):

shige (approved):
👍

satococoa (approved):
Thank you!

📝 
- The `agentActivities.totalDurationMs` is a `numeric` type.
- A `numeric` type value is treated as a `string` in TypeScript.
- Therefore, we need to convert it to number.
- Fetched value here won't be greater than `Number.MAX_SAFE_INTEGER`, this conversion will be safe.

gentamura:
Thanks 🚀

PR #254, opened 2024-12-18
[feat] Add flow retry capability with state preservation

toyamarinyon:
https://github.com/user-attachments/assets/29d6136b-87bc-4847-af8f-518a09db7931

## Changes
- Add retry button to failed execution steps
- Implement execution snapshot system to preserve graph state
- Add retry flow execution functionality that maintains execution history
- Refactor execution state management for better reliability
- Optimize blob listing with folded mode

## Testing

> [!NOTE]
> It is difficult to artificially cause an error and then succeed with a further retry, so it is OK if the normal case can be confirmed.
 
### Execute a node on the playground

- [ ] Visit `/p/:agentId`
- [ ] Open playground
- [ ] Put on node
- [ ] Generate text with Text generator node

### Execute a flow on the playground
- [ ] Visit `/p/:agentId`
- [ ] Open playground
- [ ] Put on nodes
- [ ] Connect the nodes
- [ ] Execute flow with connected nodes

### DIFF
--- a/app/(playground)/p/[agentId]/components/header.tsx
+++ b/app/(playground)/p/[agentId]/components/header.tsx
@@ -1,15 +1,31 @@
 import { GiselleLogo } from "@/components/giselle-logo";
 import Link from "next/link";
-import type { ReactNode } from "react";
+import type { ButtonHTMLAttributes, DetailedHTMLProps, ReactNode } from "react";
 import { useExecution } from "../contexts/execution";
 import { useGraph } from "../contexts/graph";
 import {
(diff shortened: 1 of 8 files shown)


toyamarinyon:
https://github.com/user-attachments/assets/ed526e1b-4b48-484e-85af-abc9bc77db59

I was able to record a stream of Overloaded errors in the Preview environment, which were resolved by re-running.
The output result of the re-run is not displayed, which will be addressed separately.

shige (approved):
LGTM! ♻️

PR #258, opened 2024-12-19
Consolidate duplicate instrumentation code between node and flow execution

toyamarinyon:
I combined overlapping parts in the processing of node execution and flow execution, as there was instrumentation code scattered around.
Since the functionality hasn't changed, it should work the same way as before.

#### Execute a node on the playground

- [ ] Visit `/p/:agentId`
- [ ] Open playground
- [ ] Put on node
- [ ] Generate text with Text generator node

#### Execute a flow on the playground
- [ ] Visit `/p/:agentId`
- [ ] Open playground
- [ ] Put on nodes
- [ ] Connect the nodes
- [ ] Execute flow with connected nodes


shige (approved):
LGTM! ♻️

PR #264, opened 2024-12-20
Measure  usage of Unstructured in playground v2

### LINKED ISSUE #259: Measure usage of Unstructured in playground v2
## Task description

We need to measure usage of Unstructured because its pricing model is related to the page of the PDF uploaded.
This means the total cost can easily doubles if user uploaded a PDF twice in page numbers.

## Background

- We're gathering usage info to optimize plan price and user experience

## TODO (Optional)

- [x] add instrumentation to the API call to Unstructured

## Additional notes (Optional)

- in beta-proto: https://github.com/giselles-ai/giselle/issues/198

Rindrics:
## Summary

Count API call for Unstructured with number of PDF pages processed

## Related Issue

this PR closes #259

## Changes

- added implementation to  API call for Unstructured in playground v2
  - removed instrumentation-related codes from action.ts to clarify business logic
- updated the corresponding code in `prev/beta-proto` agent to resolve build error
  - but the code is not refactored because it won't be used

## Testing

metrics below are arrived to the o11y backend ✅ 
- number of pages（`numPages`）
- request count
- type of API (`strategy`)

image

## Additional Inoformation

the total cost, `unit price` * `number of pages`, will be calculated on the analytics platform (not in this app)


package.json
```diff
@@ -80,6 +80,7 @@
 		"next-auth": "^5.0.0-beta.20",
 		"next-themes": "0.3.0",
 		"openai": "4.64.0",
+		"pdf-lib": "1.17.1",
```
shige, line 83: IMO:

The file size is large, and I’m hesitant to introduce it solely for the purpose of counting the number of pages in the PDF.

https://www.npmjs.com/package/pdf-lib

lib/opentelemetry/wrapper.ts
```diff
@@ -148,19 +151,30 @@ export const VercelBlobOperation = {
...
 
+interface UnstructuredOptions {
+	strategy: Strategy;
+	pdf: PDFDocument;
+}
+
+type ServiceOptions = UnstructuredOptions | VercelBlobOperationType | undefined;
+
+function getNumPages(pdf: PDFDocument) {
+	return pdf.getPages().length;
```
shige, line 162: @Rindrics 

Isn't information like the number of pages in a PDF or cost calculation references included in the response from the Unstructured API, without using pdf-lib?

shige (commented):
@Rindrics I asked a question and shared my thoughts in the comments. ✍️

lib/opentelemetry/wrapper.ts
rindrics: (Although I should have written this as PR author comment...,) yes, [`PartitionResponse`](https://github.com/Unstructured-IO/unstructured-js-client/blob/a8c02970197c73f37a15e337c60f4bbdfc361aa1/src/sdk/models/operations/partition.ts#L14-L31) does not provide it.

Firstly I attempted to obtain PDF page from the response, but found that each element in `elements` field was text block, not a page.
Thus calculating the length of `elements` will return the number of text elements, not page.

package.json
rindrics: thanks, I'll consider another approach simply obtaining number  of pages of the given PDF.

rindrics: We can access number of pages via `/Count` metadata in  the file

ref: Specification of metadata about page number can seen "7.7.3.2 Page Tree Nodes" in [PDF specification document](https://opensource.adobe.com/dc-acrobat-sdk-docs/pdfstandards/PDF32000_2008.pdf)

rindrics: found out that we need to:
- consider line breaks inside metadata tag
- ignore metadata of child tree
- consider non-compliant files which do not have a /Pages field

rindrics: [pdf-page-counter](https://www.npmjs.com/package/pdf-page-counter) (4.68 MB in size) might be a good trade-off between redevelopment (with tests!!) and additional dependencies.

rindrics: >  [pdf-page-counter](https://www.npmjs.com/package/pdf-page-counter) (4.68 MB in size) might be a good trade-off between redevelopment (with tests!!) and additional dependencies.

I think this is reasonable for our situation. Do you have any comments about this? @shige

shige: @Rindrics 

The approach of handling only the count for PDFs seems promising! 👍 

However, for the following reasons, I think it might be better to implement it directly using [pdfjs-dist](https://www.npmjs.com/package/pdfjs-dist):  

- [pdf-page-counter](https://www.npmjs.com/package/pdf-page-counter) is no longer maintained  
- It uses CommonJS

rindrics: The core code in pdf-counter is small: https://gitlab.com/decebal2dac/pdf-page-counter/-/blob/master/lib/pdf-page-counter.js?ref_type=heads

rindrics: > it might be better to implement it directly using [pdfjs-dist](https://www.npmjs.com/package/pdfjs-dist)

Thanks, I will! 🚀  @shige

shige:
Please allow me to close this for the following two reasons: @Rindrics 

- The base code has become outdated  
- We are no longer using unstructured for the time being

PR #278, opened 2024-12-26
Allow MeasurementScope and isRoute06User to be passed as parameters to withMeasurement

toyamarinyon:
The MeasurementScope and isRoute06User values can now be passed directly as parameters to withMeasurement. Previously, these values had to be retrieved from Supabase or the database within withMeasurement itself. This change allows the values to be obtained from the execution context even when there is no logged-in user, such as in webhook scenarios.


toyamarinyon:
@Rindrics I have changed the instrumentation part in this Pull request, please review.

lib/opentelemetry/wrapper.ts
```diff
@@ -1,8 +1,10 @@
 import { getCurrentMeasurementScope, isRoute06User } from "@/app/(auth)/lib";
+import { db } from "@/drizzle";
 import { waitUntil } from "@vercel/functions";
 import type { LanguageModelUsage } from "ai";
 import type { LanguageModelV1 } from "ai";
 import type { Strategy } from "unstructured-client/sdk/models/shared";
+import type { AgentId } from "../../app/(playground)/p/[agentId]/types";
```
shige: 📝 IMO:

I felt that retrieving data from the database in `lib/opentelemetry/wrapper.ts` is a bit suboptimal, but I think it's better than being unable to determine `team.type` via the current webhook setup. So, for now, I think this approach is fine.

Besides, methods like `isRoute06User` are already being imported.

shige (approved):
LGTM 👍

lib/opentelemetry/wrapper.ts
toyamarinyon: Agreed. In the future, we would like to have a function called `setMeasurementScope` that can be set on the application side as required (preferably in a common process such as middleware).

shige: Nice! 👍

Rindrics (approved):
thx!

PR #281, opened 2024-12-27
fix(execution): Update execution state during flow processing

toyamarinyon:
Modify flow execution to maintain current state by updating the initialExecution reference when execution changes. This ensures the execution context remains synchronized throughout the flow processing.

- Add onExecutionChange callback to update both state and reference
- Change initialExecution from const to let to allow updates


shige (approved):
LGTM 👍

PR #288, opened 2025-01-06
fix(components): Add key prop to text panel TabsContent

toyamarinyon:
Adds missing key prop to TabsContent component in properties panel when rendering text node content. This prevents React re-render issues and improves component stability when switching between different text nodes.


shige (approved):
LGTM 👍

PR #246, opened 2024-12-17
Send user seat meter event

satococoa:
## Summary
Add functionality to report user seat usage to Stripe at the end of each billing cycle.
The reported seat count is based on the latest team member count at billing time, regardless of fluctuations during the billing cycle. This simplifies our usage reporting by using a point-in-time measurement rather than tracking historical changes.

> [!NOTE]
> Invoice must be in draft status to report usage
> 
> - If the invoice is already finalized, we cannot report the usage
> - Consider adjusting the invoice finalization grace period in Stripe if needed

See code comments in `handleSubscriptionCycleInvoice` method.

```mermaid
sequenceDiagram
    participant Stripe
    participant Webhook as Webhook Handler
    participant Handler as handleSubscriptionCycleInvoice
    participant Reporter as reportUserSeatUsage
    participant DB
    
    Stripe->>Webhook: invoice.created event
    Note over Webhook: Check if billing_reason is<br/>subscription_cycle
    
    Webhook->>Handler: Handle draft invoice
    Note over Handler: Validate invoice status<br/>Extract subscription & customer info
    
    Handler->>Reporter: Report usage with:<br/>- subscriptionId<br/>- customerId<br/>- periodEnd
    
    Reporter->>DB: Query current team members
    DB-->>Reporter: Return member list
    
    Reporter->>Stripe: Create meter event
    Note over Reporter: Use timestamp with<br/>5 second buffer
    
    Reporter->>DB: Save usage report
    Note over DB: Store:<br/>- Team ID<br/>- Member list<br/>- Meter event ID<br/>- Timestamp
    
    Reporter-->>Handler: Complete
    Handler-->>Webhook: Complete
    Webhook-->>Stripe: 200 OK
```

## Changes
- Add new database table user_seat_usage_reports
  - Store team member snapshots for audit purposes
  - Track Stripe meter event IDs and timestamps
- Add `invoice.created` event handling
  - Implement handleSubscriptionCycleInvoice for subscription cycle billing
- Add user seat reporting logic
  - Implement reportUserSeatUsage function to report current team size
  - Save usage reports to database for tracking

## Testing
It appears that Stripe's test clock doesn't handle meter event timestamps as a test use case.
I've tested this behavior using a dummy timestamp.

```diff
diff --git a/services/usage-based-billing/user-seat.ts b/services/usage-based-billing/user-seat.ts
index 4bb4c3f..1f54545 100644
--- a/services/usage-based-billing/user-seat.ts
[...]

PR #294, opened 2025-01-07
Change user seat reporting logic

satococoa:
## Summary
Previously, user seat reporting was handled on the `invoice.created` webhook event. This approach had the following downsides:

- Reporting needed to occur while the invoice was in draft status (default: 1 hour).
- When a subscription is deleted, both `invoice.created` and `customer.subscription.deleted` events are emitted in an arbitrary order. For example, if `customer.subscription.deleted` is emitted before `invoice.created`, user seat reporting might not work correctly.

This pull request introduces a simpler logic:

- When a subscription starts or renews, send the current user seat usage as a snapshot value.
- After sending the snapshot value, delta values can be sent (including negative deltas).

Additionally, the `reportUserSeatUsage` method is designed to be idempotent, meaning it is not strongly dependent on the webhook receive timing. This ensures that user seat data will not be duplicated or inconsistent, even if events are received out of order or webhooks are retried. 

## Related Issue
- https://github.com/giselles-ai/giselle/pull/246

## Changes
- Add value (integer) and is_delta (boolean) columns to the `user_seat_usage_reports` table.
  - The value sent to Stripe will not always equal `userDbIdList.length`, therefore it is recommended to store these values.
- Add a `reportUserSeatUsage` function that takes the current user seat count reporting.
- Call the `reportUserSeatUsage` function when a new subscription is created, when an existing subscription is updated, and when a user is added or removed from a team.

## Testing
- [x] Create new Pro team, user seat is reported as 1.
- [x] Add team member, user seat is reported as +1 delta.
- [x] Add team member, user seat is reported as +1 delta.
- [x] Remove team member, user seat is reported as -1 delta.
- [x] (with test clock) update subscription, user seat is reported as 2.

### TODO
- [x] Run migration on Preview
- [ ] Run migration on Production


shige (approved):
LGTM! 💺

toyamarinyon (approved):

satococoa:
I'll fix conflicts.

satococoa:
📝 I'll run migration on Preview and do my OAT tomorrow.

satococoa:
>  Create new Pro team, user seat is reported as 1.

スクリーンショット 2025-01-08 10 57 09

Failure ❌ 
Reported twice.
I'll fix this case.

satococoa:
Added record lock to avoid race condition.
I'll do OAT again.

If it's okay, I'll re-request review about the fix.

satococoa:
I found a race condition https://github.com/giselles-ai/giselle/pull/294/#issuecomment-2576554007 and fixed in these commits: 
- https://github.com/giselles-ai/giselle/pull/294/commits/09ba06165e52956ec1f2d9cad058277c37f7e7de
- https://github.com/giselles-ai/giselle/pull/294/commits/a5aa13ecdeef5a56d5e7317b014f6e3ed42660c3

Could you review these commits? > @toyamarinyon @shige

shige (approved):
LGTM! 🐎

toyamarinyon (approved):

satococoa:
Thank you for reviews.

@shige 
Could you run migration and deploy to production?

shige:
I'll do it now!

shige:
@satococoa I did the following. 🚀 

> * [x]  Run migration on Production

`bun drizzle-kit migrate`

PR #295, opened 2025-01-07
Align property names with corresponding table column names

satococoa:
## Summary
The `agentTimeUsageReports` and `userSeatUsageReports` tables have a `created_at` column, but the corresponding property name in TypeScript is `timestamp`.
Because this inconsistency increases cognitive load, this pull request will align the property names.

This change does not require database migrations.


shige (approved):
LGTM 👍

toyamarinyon (approved):

PR #296, opened 2025-01-07
feat(billing): Implement last month usage charging on subscription cancellation

gentamura:
## Summary
Implement proper final billing handling when a subscription is cancelled, ensuring accurate billing for the last period of usage and maintaining data consistency.

## Related issue
NONE

## Changes
- Added handleInvoiceCreation module to manage invoice processing
- Added pay function for proper invoice state transitions
- Added subscription data synchronisation after cancellation

## Testing
- Subscription cancellation flow tested
- Verified correct invoice state transitions (draft → paid)
- Confirmed database synchronisation after subscription cancellation

## Other information
- Billing state transitions follow Stripe's documentation: https://support.stripe.com/questions/invoice-states
- Test clock simulation can be enabled for development testing

- Adding the agent time charge manually is both time-consuming and expensive, so it can be done in the stripe cli as follows. As this is only a registration to stripe, the db is not updated, so the ui usage of the application remains unchanged.

```
stripe billing meter_events create \
  --event-name=agent_time_charge \
  -d ‘payload[value]’=256 \
  -d ‘payload[stripe_customer_id]’={{customer_id}}
```


app/webhooks/stripe/route.ts
```diff
@@ -82,33 +83,20 @@ export async function POST(req: Request) {
 					);
 				}
 				await handleSubscriptionCancellation(event.data.object);
+				await upsertSubscription(event.data.object.id);
```
gentamura, line 95: :memo: When a subscription expires and the subscription is cancelled, `customer.subscription.updated` does not fire, so the above method is executed on `customer.subscription.deleted` and the database is updated appropriately.

app/webhooks/stripe/route.ts
```diff
@@ -82,33 +83,20 @@ export async function POST(req: Request) {
...
-						);
-						break;
-					}
-				}
+				await handleInvoiceCreation(event.data.object);
 
+				// TODO: This block will be removed in the other issue.
 				if (event.data.object.billing_reason === "subscription_cycle") {
 					await handleSubscriptionCycleInvoice(event.data.object);
 				}
```
gentamura, line 97: :memo: This block will be removed in the other issue. We plan to add user seats in other ways.

services/usage-based-billing/user-seat.ts
gentamura, line 1: :memo: Added to use the simulation function on the Stripe admin panel. We wondered whether to make it an environment variable, but decided to define it within a function as it is not a constant that is used in multiple places.

app/webhooks/stripe/handle-invoice-creation.ts
```diff
@@ -0,0 +1,30 @@
...
+	if (subscription.status !== "canceled") {
+		return;
+	}
+
+	await finalizeAndPayInvoice(invoice.id);
+}
+
+async function finalizeAndPayInvoice(invoiceId: string) {
+	try {
+		await stripe.invoices.finalizeInvoice(invoiceId);
```
gentamura: :memo: Update from `Draft` to `Open` with this method.

ref: https://support.stripe.com/questions/invoice-states?locale=en-US

app/webhooks/stripe/handle-invoice-creation.ts
```diff
@@ -0,0 +1,30 @@
...
+		await stripe.invoices.pay(invoiceId);
```
gentamura: :memo: Update from `Open` to `Paid` with this method.

ref: https://support.stripe.com/questions/invoice-states?locale=en-US

app/webhooks/stripe/route.ts
```diff
@@ -82,29 +83,15 @@ export async function POST(req: Request) {
...
```
gentamura: :memo: This block will be removed in the other issue. We plan to add user seats in other ways.

ref: https://github.com/giselles-ai/giselle/pull/294

gentamura:
@satococoa review, please :pray:

app/webhooks/stripe/handle-invoice-creation.ts
satococoa: Can we use `auto_advance` parameter here?
If we can, we would skip to call invoices.pay, invoices.send..., or so manually.
Stripe would care them.

https://docs.stripe.com/api/invoices/finalize?api-version=2024-11-20.acacia#finalize_invoice-auto_advance

services/usage-based-billing/user-seat.ts
satococoa, line 1: Thank you, but you do not have to care about this in this pull request.
Because I changed user sear reporting logic in https://github.com/giselles-ai/giselle/pull/294.

satococoa (commented):
Thank you!
Could you check my comments, please?

app/webhooks/stripe/handle-invoice-creation.ts
gentamura: @satococoa 

I will keep you posted on the progress along the way.

I ran it with `await stripe.invoices.finalizeInvoice(invoice.id, { auto_advance: true })` and the flow was as follows.

- Simulation proceeds until the end of the period and the invoice is updated from draft to open for the metered invoice.
- automatic collection of the invoice is executed one hour after the invoice is issued, so the simulation proceeds to one hour later
- Assumed to be paid at this time, but remained open previously and payment is not completed

Continue to check.

gentamura: @satococoa 
It seems that running auto_advance: true in finalizeInvoice after cancelling a subscription is not working correctly (it works like auto_advance: false), so I also checked with Stripe support and found the following method to run. The following method is called

https://github.com/giselles-ai/giselle/blob/9ba3e19835db3013037eb765f7fe8bed609f5506/app/webhooks/stripe/handle-invoice-creation.ts#L19

If the subscription is cancelled and there are usage fees remaining, a draft invoice for the usage fees will be created at the end of the subscription period.
Based on this invoice, the payment method is executed to complete the billing process and execute the actual payment.

gentamura:
@satococoa review, please :pray:

app/webhooks/stripe/handle-invoice-creation.ts
```diff
@@ -0,0 +1,20 @@
...
+
+	const subscription = await stripe.subscriptions.retrieve(
+		invoice.subscription,
+	);
+
+	if (subscription.status !== "canceled") {
+		return;
+	}
+
+	await stripe.invoices.pay(invoice.id);
```
satococoa: Could you add a comment explaining why we need to call `pay` when a subscription is canceled?

For example:
When a subscription is canceled, we should charge for usage-based billing from the previous billing cycle. The final invoice, which includes these charges, will be automatically created but will not be processed for payment. Therefore, we need to handle this case manually.

satococoa (approved):
Thank you! 
Now it's very clear and straightforward! 👍

app/webhooks/stripe/handle-invoice-creation.ts
gentamura: Added https://github.com/giselles-ai/giselle/pull/296/commits/0ff564170fead58f3a5030a34ca364a312a448f8 👍

gentamura:
@shige @toyamarinyon review, please :pray:

shige (approved):
LGTM! 💳

gentamura:
Thank you for reviewers 🚀

PR #315, opened 2025-01-17
Add permissions to the GitHub Actions workflow

masutaka:
I have already changed the `GITHUB_TOKEN` permissions to read-only at the organizational level. This follows the [Principle of least privilege](https://en.wikipedia.org/wiki/Principle_of_least_privilege).

In this pull request, I will make changes to workflows that should clearly be granted permissions.

Since all the tests in this pull request have passed, I think it is ok. If permissions-related problems occur in the future, we can deal with them on a case-by-case basis.

### DIFF
--- a/.github/workflows/license.yml
+++ b/.github/workflows/license.yml
@@ -23,6 +23,8 @@ jobs:
     timeout-minutes: 10
     env:
       LICENSE_REPORT: docs/packages-license.md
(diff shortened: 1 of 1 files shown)


shige (approved):
LGTM 🛡️

toyamarinyon (approved):

PR #356, opened 2025-02-06
Modify GitHub webhook event handling to gracefully handle null commands

satococoa:
## Summary
If the comment body does not include call sign, there is nothing to do.
In this case, handler returns 400 error now, but handling as success is better.


toyamarinyon (approved):

shige (approved):
LGTM 👍

PR #374, opened 2025-02-18
chore(deps): Upgrade Turbo from 2.3.3 to 2.4.2 and remove it from catalog

toyamarinyon:
Update Turbo dependency and remove it from catalog version control to use explicit versioning. This ensures consistent builds across environments.

### Why

I want to use [turbo-ignore](https://github.com/vercel/turborepo/blob/24f7329440446288bf662997821e8585a2ddb1bb/packages/turbo-ignore/README.md) to [skip unnecessary build steps on Vercel.](https://vercel.com/docs/monorepos/turborepo#ignoring-unchanged-builds) However, when using Turborepo with catalog version control, it was causing errors. Therefore, I've removed the catalog version control and switched to explicit versioning to resolve these issues.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
 	"devDependencies": {
 		"@biomejs/biome": "catalog:",
 		"@turbo/gen": "catalog:",
-		"turbo": "catalog:",
+		"turbo": "2.4.2",
 		"typescript": "catalog:"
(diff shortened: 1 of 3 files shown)


shige (approved):
LGTM 🆙

PR #375, opened 2025-02-18
feat(middleware): Add production access restriction

toyamarinyon:
Block access to playground app in production environment by returning 404 responses, while allowing normal access in development and staging(required Vercel login to Edge team).

This change improves security by preventing unauthorized access to the playground environment when deployed to production.


shige (approved):
LGTM! ⛔

PR #381, opened 2025-02-18
Refactor GitHub webhook integration handling and improve error tracking

### LINKED ISSUE #377: Implement Sentry notifications for GitHub authorization errors
### Self Checks

- [x] This is only for bug report, if you would like to ask a question, please head to [Discussions](../discussions/categories/q-a).
- [x] I have searched for existing issues [search for existing issues](../issues), including closed ones.

### Cloud or Self Hosted

Cloud

### Version (only for Self Hosted)

_No response_

### Steps to reproduce

1. **Setup:**
   - Find a repository that has already integrated with Giselle
   - Temporarily remove `pull_request` read/write permissions from GitHub App settings
   
2. **Test:**
   - Attempt to trigger Giselle from a pull request comment in the integrated repository

3. **Current Result:**
   - Integration fails due to insufficient permissions
   - Error is not properly tracked in monitoring system

### Expected Behavior

- When GitHub API returns 403 Forbidden due to permission issues:
  - Error is captured and sent to Sentry
  - Error context includes:
    - Repository full name
    - Installation ID
    - Error timestamp
    - Specific GitHub API endpoint that failed
- Development team receives immediate notification via Sentry
- Error is tagged appropriately for easy filtering

### Actual Behavior

- Permission errors are not captured by Sentry
- Debugging requires manual investigation

### Additional Context

_No response_

satococoa:
closes #377

## Summary
- Change Promise.all -> Promise.allSettled
- If error is occurred in integration flow, send notification to Senty
- Add log for notifyWorkflowError

### DIFF
--- a/apps/studio.giselles.ai/app/webhooks/github/handle_event.ts
+++ b/apps/studio.giselles.ai/app/webhooks/github/handle_event.ts
@@ -1,4 +1,4 @@
-import { db } from "@/drizzle";
+import { db, type githubIntegrationSettings } from "@/drizzle";
 import { saveAgentActivity } from "@/services/agents/activities";
 import { reportAgentTimeUsage } from "@/services/usage-based-billing";
 import { executeStep } from "@giselles-ai/lib/execution";
@@ -9,8 +9,9 @@ import {
 } from "@giselles-ai/lib/utils";
 import type { Execution, Graph } from "@giselles-ai/types";
 import type { Octokit } from "@octokit/core";
+import type { EmitterWebhookEvent } from "@octokit/webhooks";
 import { waitUntil } from "@vercel/functions";
-import { parseCommand } from "./command";
+import { type Command, parseCommand } from "./command";
 import {
 	assertIssueCommentEvent,
 	createOctokit,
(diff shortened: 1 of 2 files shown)


shige (approved):
LGTM 🤾

apps/studio.giselles.ai/app/webhooks/github/handle_event.ts
```diff
@@ -116,109 +118,138 @@ async function handleIssueComment(
 	);
 
 	waitUntil(
-		Promise.all(
+		Promise.allSettled(
```
toyamarinyon: Why use allSettled?

satococoa: @toyamarinyon 

With Promise.all, if any promise rejects, the entire Promise.all rejects immediately, and the `waitUntil` may consider it “done” prematurely.
I think this would cause remaining tasks to be terminated before they finish.

In contrast, Promise.allSettled waits for all tasks to complete (either fulfilled or rejected) before settling, which ensures that every task gets fully executed.

toyamarinyon: @satococoa Thanks for being careful about error handling! However, in this case, `Promise.all` would actually be sufficient since we're already handling errors comprehensively within each promise:

```typescript
integrationSettings.map(async (integrationSetting) => {
    try {
        await executeIntegrationFlow(...)
    } catch (e: unknown) {
        console.error(...);
        captureException(e, {...});
    }
});

Comment (satococoa) [reply apps/studio.giselles.ai/app/webhooks/github/handle_event.ts]: 💡 oh, make sense. You're right.

I'll change to use `Promise.all`

Comment (satococoa) [inline apps/studio.giselles.ai/app/webhooks/github/handle_event.ts]:
```suggestion
		Promise.all(
```

toyamarinyon (approved):

PR #386, opened 2025-02-20
[playground] Support multi outputs

toyamarinyon:
Because there's a prototype, a detailed review isn't necessary.

> [!NOTE]
> There are no changes to studio.giselles.ai


internal-packages/workflow-designer-ui/src/editor/index.tsx
```diff
@@ -189,43 +223,80 @@ function NodeCanvas() {
...
-											model: "gemini-2.0-flash-exp",
-											topP: 1.0,
-											temperature: 0.7,
-											searchGrounding: false,
+										type: "action",
+										content: {
+											type: "textGeneration",
+											llm: {
+												provider: "google",
+												model: "gemini-1.5-flash-latest",
```
shige, line 283: @toyamarinyon Gemini 2.0 Flash has already reached GA, but if there's a reason you're still using 1.5, could you let me know?

`gemini-2.0-flash-001`

https://sdk.vercel.ai/providers/ai-sdk-providers/google-generative-ai#model-capabilities

image

shige (approved):
LGTM 👍 👍

internal-packages/workflow-designer-ui/src/editor/index.tsx
toyamarinyon: @shige Thanks for pointing this out! You're absolutely right - there's no specific reason for using Gemini 1.5 Flash over 2.0 Flash. I'll update the model to `gemini-2.0-flash-001` since it's GA and is the more capable model.

shige: Understood. That’s cool! 👍

PR #387, opened 2025-02-20
style(ui): Rename color variables for consistency

toyamarinyon:
- Standardize color naming convention (black-40 -> black-400)
- Add white-900 color variable
- Update color references across components
- Add pop-pop animation theme variables
- Remove font-serif class from generation view

Note: No breaking changes as this is purely a styling update


shige (approved):
LGTM 🎨

PR #389, opened 2025-02-21
Update feb21

toyamarinyon:
- [x] Add generate text short cut
- [x] Set viewport to storage

Because there's a prototype, a detailed review isn't necessary.

> [!Note]
>
> There are no changes to studio.giselles.ai


shige (approved):
LGTM 👍

PR #402, opened 2025-02-26
Develop /settings/team/usage

okky-hi:
## Summary

Edited and added related components and files to reproduce the Team Usage page design.

## Changes

- Add font
- Add color
- Create a new page
  - `apps/studio.giselles.ai/app/(main)/settings/team/usage/page.tsx`
    - Display NotFound if the SETTING_V2 flag is not enabled
- Create new components to be used exclusively in settings V2
- Recreate the design for the TeamUsage page and related components

## Testing

Open[ Preview and /settings/team/usage](https://giselle-git-fork-okky-hi-feature-settings-team-usage-r06-edge.vercel.app/settings/team/usage) , on/off settings-v2 flag


apps/studio.giselles.ai/app/(main)/layout.tsx
```diff
@@ -7,8 +9,15 @@ import type { ReactNode } from "react";
 import { Nav } from "./nav";
 
 export default async function Layout({ children }: { children: ReactNode }) {
+	const settingsV2Mode = await settingsV2Flag();
```
satococoa: Could you please leave all pages/components other than `/settings` as they are?
The background color is changed in the new design, but that is not related to the `settingsV2` implementation.

apps/studio.giselles.ai/app/layout.tsx
```diff
@@ -32,16 +34,23 @@ export const metadata: Metadata = {
...
+
 	return (
 		<html lang="en" suppressHydrationWarning>
 			<GoogleTagManager gtmId={process.env.GTM_ID ?? ""} />
 			<PHProvider>
-				<body className={`${rosart.variable} font-sans`}>
+				<body
+					className={cn(
+						`${rosart.variable}  font-sans`,
+						settingsV2Mode ? `${hubot.variable} ${geist.variable}` : "",
```
satococoa: I believe that adding `${hubot.variable} ${geist.variable}` would not affect the existing pages.
Therefore, you don't need to use the `settingsV2Flag` condition here.

This feature isn't exclusive to the `settingsV2` implementation and should be available globally.

apps/studio.giselles.ai/vendor/fonts/alternative.ts
```diff
@@ -1,8 +1,18 @@
-import { Roboto_Serif } from "next/font/google";
+import { Geist, Hubot_Sans, Roboto_Serif } from "next/font/google";
```
satococoa, line 1: Could you please add these fonts to `vendor/fonts/rosart.ts` also?

In the production environment, `vendor/fonts/rosart.ts` is injected as `./fonts` instead of this file.
Reference: https://github.com/giselles-ai/giselle/blob/main/apps/studio.giselles.ai/prepare-font.js

Additionally, I would appreciate if you could rename the file from `vendor/fonts/rosart.ts` to `vendor/fonts/production.ts` for clarity.

satococoa (changes requested):
Thank you very much for your contribution!

I would like to request that you reconsider how the feature flag is being used. Could you please review the implementation approach for the feature flag?

Here is my thought:

## Problem

In this Pull Request, `settingsV2Flag()` is called multiple times in various components, which leads to a large number of places to edit when the feature flag is eventually removed. Specifically, files like `layout.tsx`, `card.tsx`, `agent-time-charge.tsx`, and `agent-usage-table.tsx` have many conditional statements (`if (settingsV2Mode) ... else ...`) and class name switches, all of which must be cleaned up when the flag is no longer needed.

## Direction

By **handling `settingsV2Flag()` at a higher level—such as at the layout or page level—** you can drastically reduce the workload of removing the feature flag later. Once the V2 feature is ready to be fully released, you can simply delete the “V1 layout” and remove a single conditional check in one place.

## Example

### Splitting layout files

For example, you can modify `apps/studio.giselles.ai/app/(main)/settings/layout.tsx` like this:

```tsx
import { settingsV2Flag } from "@/flags";
import type { ReactNode } from "react";
import { SettingsLayoutV1 } from "./settings-layout-v1";
import { SettingsLayoutV2 } from "./settings-layout-v2";

export default async function SettingsLayout({ children }: { children: ReactNode }) {
  const isV2 = await settingsV2Flag();
  
  if (isV2) {
    // Layout for V2
    return <SettingsLayoutV2>{children}</SettingsLayoutV2>;
  }
  // Layout for V1
  return <SettingsLayoutV1>{children}</SettingsLayoutV1>;
}
```

- `SettingsLayoutV1` is the existing layout (the V1 implementation).
- `SettingsLayoutV2` is the new layout that uses the updated design or classes (such as `bg-black-850` from this PR).

By doing so, you only call `settingsV2Flag()` once in the `layout.tsx`.

### Reducing if-checks in components

- Within `SettingsLayoutV2`, avoid calling `settingsV2Flag()` again; **assume the V2 look and feel is always used**.
- Likewise, the V1 components (like the existing `Card` etc.) are only referenced by `SettingsLayoutV1`, so you don’t need to mix them with V2 styling via multiple ternary operators.

## Additional Notes

- If the V2 UI differs significantly across the entire page, you could also **create a separate folder or route specifically for V2** and set up a redirect based on the flag.
- When you no longer need the flag, you can remove the `if (isV2)` in `SettingsLayout` and delete `SettingsLayoutV1`. You’ll be done with minimal refactoring, rather than having to clean up flag checks scattered throughout the codebase.

apps/studio.giselles.ai/components/v2/ui/alert.tsx
satococoa, line 1: Because this component is a app (studio.giselles.ai) global component, could you add a comment like here?:

```
FIXME: Consider integrating with apps/studio.giselles.ai/components/ui/alert.tsx when releasing setting-v2
```

apps/studio.giselles.ai/components/v2/ui/button.tsx
satococoa, line 1: Because this component is a app (studio.giselles.ai) global component, could you add a comment like here?:

```
FIXME: Consider integrating with apps/studio.giselles.ai/components/ui/button.tsx when releasing setting-v2
```

satococoa (approved):
Thank you! 👍

apps/studio.giselles.ai/components/v2/ui/alert.tsx
okky-hi: Comment addition is complete.

apps/studio.giselles.ai/components/v2/ui/button.tsx
okky-hi: Comment addition is complete.

shige (approved):
LGTM 👍

PR #408, opened 2025-02-27
Feb27a

toyamarinyon:
- [x] Rework node select intereaction

Change node select operation from click a node to double click a node. Since current interaction is a little bit complicated, I try to change the interaction.


shige (approved):
LGTM 👍

PR #411, opened 2025-02-27
Feb27b

toyamarinyon:
- [x] Show unable sources
- [x] Update text node icon
- [x] Auto detect storage backend

If there is `VERCEL` in environment variables, use vercel blob for storage backend. otherwise file storage.


shige (approved):
LGTM👍

PR #418, opened 2025-03-04
Rework gen-node select ui

toyamarinyon:
- [x] Model catalog
- [x] Model catalog UI
- [x] Writing getting started
- [x] LLM Provider from .env
- [x] Remove upload file name inference


shige (approved):
LGTM 👍

PR #431, opened 2025-03-10
GitHub Node(Prototype)

toyamarinyon:
This is a prototype of a node that allows users to retrieve any resource from GitHub using the GitHub Giselle App they have installed.

Since not all necessary features are implemented yet,  it hidden by default.

I'd like to merge this now to continue development of other features.

Although the diff is large, most of it is the schema.graphql file which contains GitHub's GraphQL API schema definition.

We could exclude this from the repository, but that would require developers to have a GitHub access token when building locally, which would complicate the build process. Therefore, we've decided to include it in the repository.


internal-packages/workflow-designer-ui/src/editor/index.tsx
```diff
@@ -166,7 +166,10 @@ function NodeCanvas() {
 				const options = {
 					ui: { position },
 				};
+				console.log(options);
```
shige:
```suggestion
```

shige: nits

internal-packages/workflow-designer-ui/src/editor/properties-panel/github-node-properties-panel/github-resource-loader.tsx
```diff
@@ -0,0 +1,177 @@
...
+
+	const handleAddUrl = useCallback(
+		async (e: FormEvent<HTMLFormElement>) => {
+			e.preventDefault();
+			setIsLoading(true);
+			const formData = new FormData(e.currentTarget);
+			const url = formData.get("url") as string;
+			const response = await client.githubUrlToObjectId({ url });
+
+			console.log(response);
```
shige:
```suggestion
```

shige (approved):
LGTM 👍

Issue #428, opened 2025-03-07
Allow Text Files Node to Process Markdown (.md) Files


shige:
## Task description
Allow Markdown files to be uploaded and processed in the Text Files Node.

## Background
Currently, the Text Files Node only supports plain text files.  Users may want to upload Markdown files to utilize the node's functionality for processing and manipulating text-based content stored in Markdown format.  This will increase the utility of the Text Files Node for users who work with Markdown documents.

> nits: 
> 
> It seems "text/markdown" accepted, but I cannot upload the [sample.md](https://github.com/user-attachments/files/19121329/sample.md).
> 
> <img width="351" alt="image" src="https://github.com/user-attachments/assets/d0de529b-250d-40a2-9072-2a752103e777" />
> 
> _Originally posted by @shige in https://github.com/giselles-ai/giselle/pull/426#discussion_r1984517379_

## TODO (Optional)
- [x] Update the Text Files Node to accept Markdown files (e.g., with `.md` or `.markdown` extensions).
- [x] Ensure that the node can correctly read and process the content of Markdown files.
- [x] Add appropriate error handling for unsupported file types.

## Additional notes (Optional)
This enhancement may require changes to file type validation and potentially adjustments to how the node parses and handles different text formats.

gentamura:
@shige 
I checked the preview at the time of [this PR](https://github.com/giselles-ai/giselle/pull/426) and it seems to have uploaded without any problems, using drag-and-drop and file selection.
I also tried local at the time of [this commit](https://github.com/giselles-ai/giselle/commit/32b9785804ba081a3eccd40ab9da6064624737d0), and there seemed to be no particular problem as well.
If there is a way to reproduce this, please let me know.

![Image](https://github.com/user-attachments/assets/f5b14556-9ba8-4f7e-bdde-d70cfa2fe04f)

Uploaded files

[sample.md](https://github.com/user-attachments/files/19175542/sample.md)
[sample.txt](https://github.com/user-attachments/files/19175543/sample.txt)

GitHub does not seem to allow uploads with the .markdown extension.

![Image](https://github.com/user-attachments/assets/573d2ab4-feeb-4d52-813d-51393b6030e1)

shige:
I will check it again in 30 minutes!

shige:
@gentamura Upon rechecking the operation, I was able to reproduce it using the following steps. 🔍 

1. Go to https://studio.giselles.ai/
2. Turn on the `new-ui` vercel flag
[...]

PR #439, opened 2025-03-12
Fix MIME Type Detection for Markdown Files

shige:
ref: https://github.com/giselles-ai/giselle/issues/428

### Problem
When uploading Markdown (.md) files, the browser sometimes returns an empty string for the MIME type instead of `text/markdown`. This causes validation to fail and prevents users from uploading Markdown files to Text nodes.

### Solution
Enhanced the file validation process to detect Markdown files by their extension when the browser doesn't provide a proper MIME type:

1. Added logic to check file extensions when MIME type is empty
2. For files with `.md` or `.markdown` extensions, explicitly set the MIME type to `text/markdown`
3. Updated both drag-and-drop validation and file selection validation

### Changes
- Modified `validateItems` function to check file extensions during drag-and-drop operations
- Updated `assertFiles` function to determine MIME type from file extension when browser returns empty type
- Improved error handling for cases where file type cannot be determined

### Testing
- Verified that Markdown files can now be uploaded successfully to Text nodes
- Confirmed that validation still works correctly for other file types
- Tested with various browsers to ensure consistent behavior

This fix ensures a better user experience when working with Markdown files in the workflow designer.

### DIFF
--- a/internal-packages/workflow-designer-ui/src/editor/properties-panel/file-node-properties-panel/file-panel.tsx
+++ b/internal-packages/workflow-designer-ui/src/editor/properties-panel/file-node-properties-panel/file-panel.tsx
@@ -103,8 +103,17 @@ export function FilePanel({ node, config }: FilePanelProps) {
 					isValid = false;
 					break;
 				}
+				// Get file extension if possible
+				let itemType = dataTransferItem.type;
+				const file = dataTransferItem.getAsFile();
+				if (!itemType && file) {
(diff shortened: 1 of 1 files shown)


shige:
@toyamarinyon @gentamura Could you review this when you have time?

toyamarinyon:
I'll review now!

toyamarinyon (approved):

shige:
Thank you for your review!  
I will merge it now 🚀 

@gentamura Please feel free to leave comments later if you have any concerns.

PR #441, opened 2025-03-12
Enable GitHub Data Injection into Giselle

toyamarinyon:
This pull request introduces the ability to inject GitHub-related information as an option into the core Giselle logic provided by the giselle-sdk.

Instead of implementing GitHub integration directly within the core Giselle logic, authentication and data retrieval are handled outside of giselle-sdk, and the resulting data is injected into Giselle as props.

This design aims to accommodate integrations with other platforms besides GitHub in the future.


shige (approved):
LGTM 👍

PR #442, opened 2025-03-12
Implement HTTP Server for Workflow Execution

toyamarinyon:
I've added an HTTP server to the workflow runtime. I'm not entirely happy with the name 'run-api', so I'll probably reconsider it at some point.

The browser runtime has some code duplication, and I'd like to unify it in the future, but I think this is fine for now.


shige (approved):
LGTM 🏃

PR #457, opened 2025-03-17
Enhanced Output Handling 

toyamarinyon:
This PR introduces a significant improvement to our node output handling system by implementing "accessors" for node outputs. This enhancement allows for more precise data flow between nodes in the workflow builder.

### Key Changes

- **Added Output Accessors**: Each node output now has an accessor property that serves as a unique identifier for that specific output type. This allows downstream nodes to reference exactly which output they need from upstream nodes.

- **Multiple Output Support**: Nodes can now produce and manage multiple distinct outputs (e.g., generated text, reasoning, sources) with each output properly identified by its accessor.

- **Output-to-Output Linking**: The generation engine now correctly maps node outputs to their corresponding generation outputs using the accessor and outputId, ensuring data flows correctly through the workflow.

### Implementation Details
- Added accessor properties to default node output definitions in the UI
- Renamed OutputBase to GenerationOutputBase and added outputId to better track output relationships
- Modified the generation resolver to use outputId for precise output targeting
- Updated the text generation completion handler to map outputs based on accessors

### Benefits
- **More Flexible Workflows**: Users can now create more sophisticated workflows by precisely controlling which outputs connect to which inputs
- **Clearer Data Flow**: The system now explicitly tracks which output data belongs to which output definition
- **Better Support for Complex Nodes**: Nodes that produce multiple outputs (like text generation with text, reasoning, and sources) are now properly supported


shige (approved):
Thanks! 🤾

PR #461, opened 2025-03-17
Gemini 2.0 Search Grounding Support

toyamarinyon:
This pull request introduces support for Search Grounding in Gemini 2.0. When Search Grounding is enabled, a "Source" output is created on the Text Generation node.

image

The "Source" output contains a JSON representation of the URLs and titles referenced during the Search Grounding process. This allows users to leverage these sources in subsequent Text Generation nodes, enabling a wider range of possibilities.

### Future Improvements
Currently, the content of the "Source" output cannot be directly viewed on the canvas. Future work will focus on providing a way to inspect the contents of the "Source" output within the canvas environment.


shige (approved):
Nice! 🔍

PR #496, opened 2025-03-21
Include team attributes to Langfuse trace to enable querying traces

### LINKED ISSUE #515: Calculate the amount of LLM API usage by user attributes
## Task description

Add user attribute metadata to Langfuse traces for tracking LLM API usage by user segments

## Background

We need better visibility into our LLM API usage patterns across different user segments and subscription tiers. By adding appropriate user attribute metadata to our Langfuse traces, we can analyze API consumption based on user types (internal vs. customer) and subscription plans (Free vs. Pro).

Rindrics:
## Summary

Add team attributes （`isProPlan` or `isInternalTeam`） to Langfuse traces to enable to resolve #515

## Related Issue

- https://github.com/giselles-ai/giselle/issues/515

## Changes

- injected team attributes using functionality introduced by https://github.com/giselles-ai/giselle/pull/503

## Testing

confirmed on preview env ✅ 
image

### DIFF
--- a/apps/studio.giselles.ai/app/workspaces/[workspaceId]/layout.tsx
+++ b/apps/studio.giselles.ai/app/workspaces/[workspaceId]/layout.tsx
@@ -40,7 +40,7 @@ export default async function Layout({
 			telemetry={{
 				metadata: {
 					isProPlan: isProPlan(currentTeam),
-					isInternalTeam: isInternalTeam(currentTeam),
+					teamType: currentTeam.type,
 				},
 			}}
 		>
--- a/apps/studio.giselles.ai/app/workspaces/[workspaceId]/layout.tsx
+++ b/apps/studio.giselles.ai/app/workspaces/[workspaceId]/layout.tsx
@@ -1,7 +1,7 @@
 import { db } from "@/drizzle";
 import { getGitHubIntegrationState } from "@/packages/lib/github";
 import { getUsageLimitsForTeam } from "@/packages/lib/usage-limits";
-import { fetchCurrentTeam } from "@/services/teams";
+import { fetchCurrentTeam, isProPlan, isInternalTeam } from "@/services/teams";
 import { WorkspaceId } from "@giselle-sdk/data-type";
 import { WorkspaceProvider } from "giselle-sdk/react";
 import { notFound } from "next/navigation";
@@ -41,6 +41,8 @@ export default async function Layout({
 			telemetry={{
 				metadata: {
 					example: "metadata",
+					isProPlan: isProPlan(currentTeam),
+					isInternalTeam: isInternalTeam(currentTeam),
 				},
 			}}
 		>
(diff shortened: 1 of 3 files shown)


apps/studio.giselles.ai/app/workspaces/[workspaceId]/layout.tsx
```diff
@@ -37,10 +37,10 @@ export default async function Layout({
 				github: gitHubIntegrationState,
 			}}
 			usageLimits={usageLimits}
-			// following settings is test. We'll replace for production.
 			telemetry={{
 				metadata: {
-					example: "metadata",
+					isProPlan: isProPlan(currentTeam),
```
Comment (shige) [inline apps/studio.giselles.ai/app/workspaces/[workspaceId]/layout.tsx:42]:
```suggestion
					teamType: currentTeam.type,
```

Comment (shige) [reply apps/studio.giselles.ai/app/workspaces/[workspaceId]/layout.tsx:42]: @Rindrics 
Using isXxx could lead to maintenance costs when the number of team types increases.
I was thinking that we might be able to distinguish them more simply like this—what do you think?

Comment (rindrics) [reply apps/studio.giselles.ai/app/workspaces/[workspaceId]/layout.tsx:42]: Thanks, your suggestion seems reasonable @shige 

The name "internal" is laced with the rule "to treat the team as 'internal' or not", which has nothing to do with the application.

Maybe we should have this rule on the business analysis platform.

The same can be said for the field `isProPlan`.

Comment (rindrics) [reply apps/studio.giselles.ai/app/workspaces/[workspaceId]/layout.tsx:42]: > The same can be said for the field isProPlan.

but at this point we don't have schema for name of the plan itself
```typescript
export function isProPlan(team: CurrentTeam) {
    return team.activeSubscriptionId != null || team.type === "internal";
}
```

considering this, adding `isProPlan` for telemetry is reasonable in practical point of view.

apps/studio.giselles.ai/app/workspaces/[workspaceId]/layout.tsx
```diff
@@ -37,10 +37,10 @@ export default async function Layout({
...
+					isInternalTeam: isInternalTeam(currentTeam),
```
Comment (shige) [inline apps/studio.giselles.ai/app/workspaces/[workspaceId]/layout.tsx]:
```suggestion
					teamType: currentTeam.type,
```

Comment (shige) [reply apps/studio.giselles.ai/app/workspaces/[workspaceId]/layout.tsx]: @Rindrics https://github.com/giselles-ai/giselle/pull/496/files#r2011207714
You're right. The part where I commented the suggested change was actually incorrect.
Currently, the plans are only Free and Pro, so keeping it as a boolean might be fine.

That said, I thought that instead of using a boolean, simply having team.type would make the code more adaptable to future changes.

toyamarinyon (approved):

shige (approved):
Thank you! 👍 

image

rindrics:
thank you for your reviews!

PR #497, opened 2025-03-22
Temporarily hide the GitHubResourceLoader node in toolbar

shige:
## Summary
Temporarily hide the GitHubResourceLoader node in the toolbar as it is not functioning properly.

## Related Issue
None.

## Changes
Just hide the GitHubResourceLoader node in toolbar.

## Testing
Please check it in playground page.


shige:
Upon checking the latest status of https://studio.giselles.ai/, it is no longer displaying, so I will close it.

PR #514, opened 2025-03-24
Add functionality to restrict heavy model usage

satococoa:
## Summary
During our launch campaign, we should prepare measures to restrict users with excessive model usage.

## TODO

- [x] migration on Preview
- [ ] migration on Production


shige (approved):
Thank you! 👍

satococoa:
@toyamarinyon cc/ @shige 

> Relying on environment variables means that any changes to these restrictions would require a redeployment. 

I share the same concern.
Let's move these settings to the database.

But I don't think we need functionality to manage records.
Given that this functionality would likely be used infrequently, I think it would be sufficient to manage records directly in the database rather than building a dedicated UI. What do you think?

After our campaign, we need to address better rate limit functionality.

satococoa:
I've changed the implementation to use the database instead of an environment variable.
This change allows us to add or remove restrictions without requiring a deployment.

shige (approved):
Thank you! ✔️

toyamarinyon (approved):
Great work on this PR\! I appreciate how you adapted the implementation from environment variables to database storage, which provides much more flexibility for managing restrictions without redeployment.

The code is clean, well-structured, and handles all the necessary cases properly. I especially like how you've set up the database relationships and indexes correctly.

This gives us a solid foundation for managing usage during the launch campaign, while allowing us to develop more comprehensive rate limiting in the future. Approved\!

satococoa:
@shige 
I've successfully done migration on Preview environment.
Could you execute migration and deploy on Production, please?

shige:
@satococoa I did it! 🚀 

```sh
pnpm drizzle-kit migrate

[✓] migrations applied successfully!  
```

Issue #498, opened 2025-03-23
Agent Name Not Updating in /agents


shige:
### Self Checks

- [x] This is only for bug report, if you would like to ask a question, please head to [Discussions](https://github.com/giselles-ai/giselle/discussions/categories/q-a).
- [x] I have searched for existing issues [search for existing issues](https://github.com/giselles-ai/giselle/issues), including closed ones.

### Cloud or Self Hosted

Cloud

### Version (only for Self Hosted)

9c0a8a350c6b1aa303de3a72f5fcf7fecae0fcb8

### Steps to reproduce

1.  Click "New Agent +".
2.  Navigate to `https://studio.giselles.ai/workspaces/wrks-YOURID` (replace `YOURID` with the actual workspace ID) and change the agent name.
3.  Navigate to `https://studio.giselles.ai/agents`.  The agent name is not updated and still shows the previous name.

https://github.com/user-attachments/assets/0e642dc1-0f5a-42e3-8885-49e0a289031c

### Expected Behavior

The agent name on the `/agents` page (Agent List) should reflect the changes made in the New Playground page (workspace edit page).  After changing the name in the workspace, the updated name should be displayed in the list.

### Actual Behavior

The agent name on the `/agents` page does not update and remains the old name after changing it on the workspace edit page.

### Additional Context

The issue occurs on the New Playground page.  The problem is specifically with the synchronization of the agent name between the workspace edit view and the agent list view.

kentaro-m:
I'll work on this.

kentaro-m:
The cause investigation and solution are as follows:

**Cause:**

When the Agent Name was changed in `/agents`, the Workspace storage was updated, but the `name` column in the `agent` table was not.

* [https://github.com/giselles-ai/giselle/blob/1ea9d6f606a17e932bee7d342477488ac5241321/packages/workflow-designer/src/react/workflow-designer-context.tsx#L151-L156](https://github.com/giselles-ai/giselle/blob/1ea9d6f606a17e932bee7d342477488ac5241321/packages/workflow-designer/src/react/workflow-designer-context.tsx#L151-L156)
[...]

PR #534, opened 2025-03-25
Apply agent name change to the /agents page when updated in the header.

kentaro-m:
## Summary

When updating the agent name from the new Playground header, the agent name is now reflected on the /agents page.
Please see the following video for detailed behavior.

https://github.com/user-attachments/assets/7eefba59-7753-4161-a131-fd290c630724

## Related Issue

https://github.com/giselles-ai/giselle/issues/498

## Changes

- Added a server action to update the agent name (name column of the agents table).
- Added a process to call the above server action in the new Playground header.

## Other Information

To prevent `internal-packages/workflow-designer-ui` from depending on Drizzle, the DB update process is now performed within `apps/studio.giselles.ai`.


internal-packages/workflow-designer-ui/src/header/component.tsx
```diff
@@ -28,6 +31,17 @@ export function Header({
 		setView(view === "editor" ? "viewer" : "editor");
 	};
 
+	const updateAgentName = (value?: string) => {
+		if (!value) {
+			return;
+		}
+
+		if (onEditableTextChange) {
```
satococoa: `updateName` should be called even if `onEditableTextChange` is not given.

internal-packages/workflow-designer-ui/src/header/component.tsx
```diff
@@ -18,8 +19,10 @@ import { SettingsPanel } from "../settings";
 
 export function Header({
 	action,
+	onEditableTextChange,
```
satococoa: I suggest renaming `onEditableTextChange` to `onWorkflowNameChange` to better reflect the domain concept. This makes the prop's purpose clearer by explicitly stating it handles workflow name changes.

internal-packages/workflow-designer-ui/src/header/component.tsx
```diff
@@ -28,6 +31,17 @@ export function Header({
 		setView(view === "editor" ? "viewer" : "editor");
 	};
 
+	const updateAgentName = (value?: string) => {
```
satococoa: I also suggest renaming `updateAgentName` to `updateWorkflowName` for consistency with the proposed `onWorkflowNameChange` prop. This creates a clearer naming pattern that reflects the component's actual purpose of editing workflow names rather than agent-related concepts.

- `Agent` is studio.giselles.ai domain specific concept.
- `Agent` is being renamed to `App` in studio.giselles.ai.
- The workflow-designer-ui package should use domain-agnostic terminology where possible.

satococoa (changes requested):
Thank you for your contribution!
I've added some comments, could you check them please?

internal-packages/workflow-designer-ui/src/header/component.tsx
kentaro-m: I fixed it. https://github.com/giselles-ai/giselle/pull/534/commits/d778de3a3e9e6dce7a59b1a4c7d02072013c4ab0

internal-packages/workflow-designer-ui/src/header/component.tsx
kentaro-m: Thanks for your suggestion.
I ranemed it. https://github.com/giselles-ai/giselle/pull/534/commits/1724e05dbf680015ef030891c09d475cd095305e https://github.com/giselles-ai/giselle/pull/534/commits/a8390a4aac82743bb572397cf53c2c3067517a72

internal-packages/workflow-designer-ui/src/header/component.tsx
kentaro-m: I understand the naming pattern and modified it with your suggestion. https://github.com/giselles-ai/giselle/pull/534/commits/1dcd6ff7630d5a3183c17f044f3c3ae4d697c72f

satococoa (approved):
great 👍  Thank you!

shige (approved):
Thank you! 🚀

PR #537, opened 2025-03-26
feat(workflow-designer-ui): Improve node input UI

satococoa:
## Changes
- Display blank `Input` on the left of generation node when no input is connected.
  - To highlight the `input` feature to users.
- Rename "Source" to "Input" in Properties panel.
  - Since the Output has a `Source` feature for search grounding, we want to prevent user confusion.
- Disable connection via dragging feature of React Flow.
  - This feature is currently not implemented in our application.

### DIFF
--- a/internal-packages/workflow-designer-ui/src/editor/node/node.tsx
+++ b/internal-packages/workflow-designer-ui/src/editor/node/node.tsx
@@ -190,6 +190,23 @@ export function NodeComponent({
 								</div>
 							</div>
 						))}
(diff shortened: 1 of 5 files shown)


shige (approved):
Thank you! 🆙 

📝 NOTE

> * Disable connection via dragging feature of React Flow.
>   
>   * This feature is currently not implemented in our application.

We plan to make improvements in the future to make the node-based UI more intuitive to use. ⛓️

Issue #509, opened 2025-03-24
Return either 402 or 406 when UsageLimitError is thrown in the http/router


satococoa:
I'd like to return either 402 or 406 when this error is thrown in the http/router. It doesn't have to be in this pull request.

_Originally posted by @toyamarinyon in https://github.com/giselles-ai/giselle/pull/495#discussion_r2009427068_

kentaro-m:
I'll work this.

PR #543, opened 2025-03-26
Add error handling for usage limit in text and image generation for giselle-engine

kentaro-m:
## Summary

Add error handling for usage limit in text and image generation for giselle-engine

## Related Issue

https://github.com/giselles-ai/giselle/issues/509

## Changes

I changed the giselle-engine text and image generation service to return HTTP Status 429 to clients when usage limit errors occur.

## Other Information

Issue #509 indicates that HTTP status codes 402 or 406 should be returned when a Usage Limit Error occurs. 

However, after I have checked again about proper HTTP status codes, it seems that returning 429 is the standard practice for rate limiting.  I would appreciate your feedback on this.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/429
https://sdk.vercel.ai/docs/advanced/rate-limiting

### DIFF
--- a/packages/giselle-engine/src/http/router.ts
+++ b/packages/giselle-engine/src/http/router.ts
@@ -194,8 +194,15 @@ export const createJsonRouters = {
 				overrideNodes: z.array(OverrideNode).optional(),
 			}),
 			handler: async ({ input }) => {
-				const result = await giselleEngine.runApi(input);
-				return new Response(result);
+				try {
+					const result = await giselleEngine.runApi(input);
+					return new Response(result);
+				} catch (error) {
+					if (error instanceof UsageLimitError) {
+						return new Response(error.message, { status: 429 });
+					}
+					throw error;
+				}
 			},
 		}),
 	generateImage: (giselleEngine: GiselleEngine) =>
--- a/packages/giselle-engine/src/http/router.ts
+++ b/packages/giselle-engine/src/http/router.ts
@@ -14,10 +14,10 @@ import {
 	WorkspaceId,
 } from "@giselle-sdk/data-type";
 import { z } from "zod";
-import { type GiselleEngine, UsageLimitError } from "../core";
+import type { GiselleEngine } from "../core";
 import type { TelemetrySettings } from "../core/generations";
 import { JsonResponse } from "../utils";
-import { createHandler } from "./create-handler";
+import { createHandler, withUsageLimitErrorHandler } from "./create-handler";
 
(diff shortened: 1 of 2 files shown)


packages/giselle-engine/src/http/router.ts
```diff
@@ -60,11 +60,18 @@ export const createJsonRouters = {
...
+					const stream = await giselleEngine.generateText(
+						input.generation,
+						input.telemetry,
+					);
+					return stream.toDataStreamResponse();
+				} catch (error) {
+					if (error instanceof UsageLimitError) {
+						return new Response(error.message, { status: 429 });
+					}
+					throw error;
```
kentaro-m: [Discussion]
Is additional error handling necessary for errors other than usage limit errors?

satococoa: Currently, handling `UsageLimitError` is sufficient. 👍

packages/giselle-engine/src/http/router.ts
```diff
@@ -195,8 +202,15 @@ export const createJsonRouters = {
 		createHandler({
```
satococoa: `runApi` above is calling generateText internally.
Could you also add an error handler to it?

Additionally, if possible, I would greatly appreciate it if you could make this error handler more modular/reusable.

satococoa (commented):
Thank you!

packages/giselle-engine/src/http/router.ts
kentaro-m: > runApi above is calling generateText internally.
> Could you also add an error handler to it?

Fixed it
https://github.com/giselles-ai/giselle/pull/543/commits/84645fdaa6e271511a35c94109040cd9644cef01

---

> Additionally, if possible, I would greatly appreciate it if you could make this error handler more modular/reusable.

I would like to implement reusable error handling like this.

```ts
const handlerWithLimit = withUsageLimitErrorHandler(
    createHandler({
        input: undefined,
        handler: async ({ input }) => {
            // logic
        },
    }),
);
```

kentaro-m: @satococoa 
I've made improvements to the error handling. Could you please review it again?
https://github.com/giselles-ai/giselle/pull/543/commits/354b91df2239b9fd8ef5e4667f9d41afbe83306b

satococoa (approved):
Thank you 👍

shige (approved):
Thank you! 🚀

PR #545, opened 2025-03-26
Improve CI workflow

toyamarinyon:
## Summary
- Optimized CI workflow to run only on main branch and PRs
- Added SDK build step before running tests
- Updated test scripts and fixed test setup

## Background
Created this PR since the current CI workflow has two problems:
1. It does not run any tests properly
2. It runs on every push and PR, which is inefficient

This PR implements a solution with a single job that runs Biome, builds the SDK, and runs tests. In the future, we should consider splitting these steps and using GitHub Actions caching to improve developer experience (DX) and speed up the workflow.

## Test plan
- Verify CI workflow runs correctly on main branch and PRs
- Confirm test runs pass with the updated configuration

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,6 +2,7 @@ name: CI
 
 on:
   push:
+    branches: [main]
   pull_request:
 
 jobs:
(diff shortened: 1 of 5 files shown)


shige (approved):
Thank you! 🆙

PR #548, opened 2025-03-27
Improve Contributing Section in README

masutaka:
This pull request improves the *Contributing* section in `README.md` to make it more inviting and easier to understand for potential contributors.

With this change, we clarified that starring the project and following it on various social media platforms are also forms of contribution, and we presented ways to do so.


shige (approved):
Thank you! 🌟

PR #529, opened 2025-03-25
Improve panel design

kaochannel154:
## Summary
Implemented multiple UI improvements to enhance visual consistency of the user interface. The focus was on adjusting panel indicators design, styling resize handles, and adding a generation button to the results area.

## Changes
* Adjusted panel indicator design
* Fixed missing style implementation for resize handles
  - Applied consistent styling across editor panels
  - Added visual feedback on hover
* Implemented new generation button in the results area
  - Added a rounded button with star icon
  - Implemented rotation animation effect on hover

## Testing
* Confirmed hover animation on the generation button works as expected

## Other Information
These UI improvements aim to enhance the user experience. In particular, the adjustment of panel indicators allows users to more clearly understand their current working context. The styling of resize handles and the animation effect of the generation button are designed to improve interface consistency and engagement.

PR #552, opened 2025-03-27
Remove hardcoded dev server port

toyamarinyon:
## Summary
- Remove hardcoded port 6180 from playground package.json to use Next.js default (3000)
- Update documentation to explain how to change port using PORT environment variable
- Update localhost URLs in documentation to reflect the default port

## Background
Originally, hardcoded dev server port (6180) was used to prevent port conflicts between different dev servers. However, this approach creates problems with collaborative tools and vibe coding environments (as noted in PR #529 where port conflicts occurred).

Following next.js conventions by using the default port (3000) and allowing override through the PORT environment variable is a better approach for:

1. Developers using vibe coding tools like Cursor, Cline, etc.
2. Following standard conventions that most developers would expect
3. Maintaining consistency with framework defaults while still allowing customization

## Test plan
- Run the dev server without specifying a port and confirm it runs on port 3000
- Test setting PORT environment variable to override the default port

### DIFF
--- a/apps/playground/README.md
+++ b/apps/playground/README.md
@@ -32,6 +32,12 @@
     pnpm turbo dev
     ```
 
+    By default, the development server runs on port 3000 (this is the standard Next.js default). If you need to use a different port, you can set the `PORT` environment variable:
+
+    ```sh
+    PORT=6180 pnpm turbo dev
+    ```
+
 ## Tasks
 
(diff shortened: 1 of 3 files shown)


shige (approved):
Thank you! 🔢 

I have confirmed that `pnpm run dev` and `pnpm run dev:studio.giselles.ai` can run simultaneously! ✅

PR #563, opened 2025-03-28
Fix Scroll Issue for Long Result Content

okky-hi:

### DIFF
--- a/internal-packages/workflow-designer-ui/src/editor/properties-panel/text-generation-node-properties-panel/generation-panel.tsx
+++ b/internal-packages/workflow-designer-ui/src/editor/properties-panel/text-generation-node-properties-panel/generation-panel.tsx
@@ -76,7 +76,7 @@ export function GenerationPanel({ node }: { node: TextGenerationNode }) {
 		return <Empty />;
 	}
 	return (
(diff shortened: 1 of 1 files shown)


internal-packages/workflow-designer-ui/src/editor/properties-panel/text-generation-node-properties-panel/generation-panel.tsx
```diff
@@ -108,7 +108,7 @@ export function GenerationPanel({ node }: { node: TextGenerationNode }) {
 					/>
 				)}
 			</div>
-			<div className="py-[4px] px-[16px] overflow-y-auto h-full">
+			<div className="flex-1 py-[4px] px-[16px] overflow-y-auto">
```
okky-hi, line 111: The parent element (line 79) is set to flex flex-col to stack its children vertically. 
By applying flex-1 to this element, the child elements expand vertically, making the height clear. 
As a result, overflow-y-auto works correctly, allowing the text to be fully scrollable without being cut off.

↓Case 1
名称未設定

↓Case 2
名称未設定3

toyamarinyon (approved):

shige (approved):
Thank you! 🚀

PR #576, opened 2025-03-31
fix: Filter active subscriptions only in fetchUsageLimits

satococoa:
This PR fixes an issue where fetchUsageLimits was not properly filtering subscription status, causing incorrect usage limit calculations.

## Changes
- Added a condition to filter only active subscriptions in the SQL join
- Uses `and` operator to combine team ID and subscription status conditions


toyamarinyon (approved):

shige (approved):
Thank you! 🛡️

Issue #486, opened 2025-03-19
Support for Specifying Recraft Style via providerOptions in generateImage


toyamarinyon:
**Problem:**

Currently, when using Recraft through Fal with the AI SDK, there doesn't seem to be a direct way to specify the desired style for image generation.  I'd like to be able to control the style of the generated image.

**Proposed Solution:**

It would be highly beneficial if the `style` parameter could be passed through the `providerOptions` of the `generateImage` function in the AI SDK.  This would allow users to leverage Recraft's styling capabilities more effectively.

**Details:**

*   The Fal Provider in the AI SDK doesn't currently expose a `style` property directly.
*   The `generateImage` function offers `providerOptions` for provider-specific configurations.
*   Passing the `style` parameter via `providerOptions` would provide a flexible and consistent way to control Recraft's image generation style.

**Relevant Documentation:**

[https://sdk.vercel.ai/docs/reference/ai-sdk-core/generate-image#provider-options](https://sdk.vercel.ai/docs/reference/ai-sdk-core/generate-image#provider-options)

**Example (Illustrative):**

```javascript
const image = await ai.generateImage({
  model: 'fal/recraft',
  prompt: 'A futuristic cityscape',
  providerOptions: {
    style: 'cyberpunk' // Example: Passing the style here
  }
});
```

**Benefits:**

*   Enhanced control over image generation style.
*   Improved integration with Recraft's features.
*   More flexible and powerful image generation capabilities within the AI SDK.

**Is there any way to achieve this currently?  If not, could this be considered as a feature request?**

ms2sato:
I would like to work on this issue.

PR #586, opened 2025-03-31
feat: add providerOptions key for params on generateImage

ms2sato:
## Summary

**example code**

```ts
await generateImage({
	context,
	generation,
	providerOptions: {
		fal: {
            		num_images: 2 // for creation 2 images
          	}
        },
});
```

image

Just wondering — If `generateImage` is specifically tied to `fal`, then the providerOptions interface could be simplified like this:

```ts
await generateImage({
	context,
	generation,
	providerOptions: {
		num_images: 2 // NO fal section
        },
});
```

## Related Issue

closed https://github.com/giselles-ai/giselle/issues/486

## Testing

`generateImage` function's argument to add `providerOptions` property at `packages/giselle-engine/src/core/index.ts:147`.

## Other Information

For example, The parameter information for fal-ai/flux/schnell can be found at [that link](https://fal.ai/models/fal-ai/flux/schnell/api#schema-input),
but verifying how the other options actually behave seems a bit challenging(num_images is easy to checking).

### DIFF
--- a/packages/giselle-engine/src/core/generations/generate-image.ts
+++ b/packages/giselle-engine/src/core/generations/generate-image.ts
@@ -32,9 +32,14 @@ import {
 	setNodeGenerationIndex,
 } from "./utils";
 
+type ProviderOptions = Parameters<
+	typeof generateImageAiSdk
+>[0]["providerOptions"];
+
 export async function generateImage(args: {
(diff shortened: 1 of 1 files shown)


shige (approved):
Thank you! 🖼️

toyamarinyon (approved):

PR #587, opened 2025-03-31
Add override nodes support to run data

toyamarinyon:
## Summary
- Adds overrideNodes array to run data schemas to store node content overrides
- Implements data modification utility to handle missing overrideNodes
- Moves override functionality from run-api to add-run for better separation of concerns
- Updates workflow data structure to apply overrides during run creation

This modification is part of implementing the "run with override parameters" feature. It provides the underlying data structures and logic needed to support a run button with override parameters in the UI.

## Test plan
- Verify runs created with overrideNodes properly store the overrides
- Ensure backward compatibility with existing runs by testing with and without overrideNodes
- Confirm node content overrides are properly applied to generation templates


shige (approved):
Thank you! 🌄

PR #600, opened 2025-04-01
Fix node layout scrolling issue

toyamarinyon:
## Summary
- Fixed layout issue where connected nodes appeared outside the viewport and users couldn't scroll to see them

## Test plan
- Verify that all nodes in workflows are visible and can be scrolled to
- Check that UI layout remains consistent across different viewport sizes


shige (approved):
Thank you! 🆙

PR #606, opened 2025-04-01
fix: replace node to text on buildGenerationMessageForTextGeneration

ms2sato:
### before
`sources` are treated as JSON node.

image

### after
`sources` are treated as Text related JSON node.

image

## Testing

For example, Input "Hello, world" in text-node and "What does  `@text/Output` mean to a programmer?" Input related text-generate-node. and generate.

### DIFF
--- a/packages/giselle-engine/src/core/generations/utils.ts
+++ b/packages/giselle-engine/src/core/generations/utils.ts
@@ -107,10 +107,11 @@ async function buildGenerationMessageForTextGeneration(
 		const replaceKeyword = `{{${sourceKeyword.nodeId}:${sourceKeyword.outputId}}}`;
 		switch (contextNode.content.type) {
 			case "text": {
-				userMessage = userMessage.replace(
-					replaceKeyword,
-					jsonContentToText(JSON.parse(contextNode.content.text)),
-				);
+				const jsonOrText = contextNode.content.text;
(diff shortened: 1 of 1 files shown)


shige (approved):
Thank you! 👏

packages/giselle-engine/src/core/generations/utils.ts
```diff
@@ -109,7 +109,7 @@ async function buildGenerationMessageForTextGeneration(
 			case "text": {
 				userMessage = userMessage.replace(
 					replaceKeyword,
-					contextNode.content.text,
+					jsonContentToText(JSON.parse(contextNode.content.text)),
```
toyamarinyon: It's difficult to understand because it's not clearly represented in the TypeScript types, but `contextNode.content.text` can be a string type, not just a `jsonContent` type.  Before parsing, please check if it's actually a `jsonContent` type. If it's not, please leave it as is (don't attempt to parse it).

I think the following code might be helpful as a reference:

https://github.com/giselles-ai/giselle/blob/e71f136fecaed452175c12f06ad62f48f709e089/packages/giselle-engine/src/core/generations/utils.ts#L89C1-L91C3

toyamarinyon (changes requested):

toyamarinyon (approved):
This is absolutely perfect! Thank you so much for your work

PR #614, opened 2025-04-02
fix(workflow-designer-ui): improve trigger change handling with safe parsing

satococoa:
Fixes this error:

```
ZodError
[
  {
    "received": "",
    "code": "invalid_enum_value",
    "options": [
      "github.issue_comment.create",
      "github.pull_request_comment.create"
    ],
    "path": [],
    "message": "Invalid enum value. Expected 'github.issue_comment.cre...
```


toyamarinyon (approved):
Since this is a response to an ongoing error, I think this is fine for now. However, I'm concerned about how long the "new" will remain. Also, I have a feeling that if the schema changes, it will become "newNew..." and so on. I would like you to use the data-mod package to migrate the data

shige (approved):
Thank you! 🏥

PR #619, opened 2025-04-02
feat(apps): update agent duplication to use workspace

gentamura:
## Summary
Refactor agent duplication functionality to use the new workspace-based system instead of the legacy graph-based implementation, while maintaining backward compatibility.

## Related Issue
N/A

## Changes
- Add copyWorkspace functionality to giselle-engine
- Remove graph-based implementation in copyAgent
- Update agent duplication to use workspace-based system
- Remove unnecessary graphUrl null check and returning clause
- Update redirect path from /p/{agentId} to /workspaces/{workspaceId}
- Add TODO comment for future graphUrl field removal
- Keep graphUrl field temporarily for backwards compatibility

## Testing
- Tested agent duplication functionality
- Verified workspace data is correctly copied
- Confirmed correct redirect to new workspace page
- Verified all existing functionality remains working

## Other Information
The graphUrl field is kept for backward compatibility but marked for future removal as it's not used in the new playground system. This change is part of the broader migration from graph-based to workspace-based implementation.


gentamura:
~:memo: Workspace can be moved, but node and flow do not seem to be copied.~

=> ✅  This is fixed.

gentamura:
@shige @toyamarinyon Code review please :pray:
Just to confirm one point, is graphUrl not used in the new playground?

shige:
### 📝 My OATs memo

It works! ✅ 

---

The only differences in workspace.json were in the id and name. 

@gentamura cc/ @toyamarinyon 

Is it acceptable if the rest of the contents, aside from id and name, are exact copies? If there are any considerations from a system perspective that should be taken into account, please let me know.

```sh
diff .storage/workspaces/wrks-AAA/workspace.json .storage/workspaces/wrks-BBB/workspace.json

2,3c2
<   "id": "wrks-AAA",
<   "name": "Copy of ",
---
>   "id": "wrks-BBB",
```

gentamura:
@shige 

> Is it acceptable if the rest of the contents, aside from id and name, are exact copies? If there are any considerations from a system perspective that should be taken into account, please let me know.

The data that make up the Workspace are the following parts.

https://github.com/giselles-ai/giselle/blob/7a08bd5a06c33309543b5b0d9f321878dfe69ec2/packages/giselle-engine/src/core/workspaces/copy-workspace.ts#L20-L28

Currently, all elements except id and name are copied as is, as expected. However, as you pointed out, we are also copying the editing state (e.g., selecting a node), so it may seem a little strange to have some panel open when the copied Workspace is accessed.

shige (approved):
Thank you! 👍 

@toyamarinyon Could you please review this when you have a moment?

📝 
If it looks like we won’t be able to merge it today, I'm thinking of going with the approach of hiding the Duplicate icon.

packages/giselle-engine/src/core/workspaces/copy-workspace.ts
toyamarinyon, line 1: It needs to copy files

https://github.com/giselles-ai/giselle/blob/fce4434583d2e7d179268c7481d217e8ef2b4d4c/packages/giselle-engine/src/core/workspaces/create-sample-workspace.ts#L146-L171

toyamarinyon (changes requested):

toyamarinyon (approved):

gentamura:
@toyamarinyon Thanks for your review 🙌

gentamura:
Thank you for reviewers 🚀

PR #627, opened 2025-04-02
Fix workflow designer UI connection handling

toyamarinyon:
## Summary
- Improve connection handling in workflow designer UI with useCallback optimization
- Add proper validation for connections by checking handle IDs instead of just nodes
- Fix error toast implementation for better error handling

## Test plan
- Test creating connections between nodes in the workflow designer
- Verify that connections between the same nodes but different handles work correctly
- Test that invalid connections are properly rejected with error messages


shige (approved):
Thank you! 🔌

PR #628, opened 2025-04-02
Improve connection validation in workflow designer

toyamarinyon:
## Summary
- Improve connection validation logic to check for duplicate connections at the handle level
- Add useCallback for handleConnect function with proper dependencies
- Refactor isValidConnection to implement more precise validation

## Test plan
- Create a workflow with multiple nodes
- Verify that connections between the same input and output handles are prevented
- Confirm that different handles between the same nodes can still be connected


shige (approved):
Thank you! 👍

PR #630, opened 2025-04-03
Improve canvas trackpad gestures for better UX

toyamarinyon:
## Summary
- Modified trackpad behavior in the canvas to match standard web applications and Figma/FigJam
- Changed interaction model so single click selects node and shows properties panel
- Double-click now only focuses the viewport on the node without affecting selection
- Two-finger trackpad scroll now pans the canvas instead of zooming (more intuitive)
- Pinch gestures still handle zooming as expected

## Test plan
- Open the workflow designer canvas in the playground
- Test single-clicking on nodes to verify selection and properties panel display
- Test double-clicking to verify viewport centering on nodes
- Test trackpad two-finger scroll to verify panning instead of zooming
- Test pinch gestures to verify zooming still works correctly

Resolves route06/giselle-division#2913


shige (approved):
Thank you! ✨

PR #638, opened 2025-04-03
Emphasize Editable Fields

okky-hi:
## Summary

We added a pencil icon, which is commonly used to indicate editability, to make it clearer that gen-nodes can be edited.

## Changes

- Added icon in button
- Styles were added along with the icons.

## Testing

1. Open [Preview](https://giselle-git-fork-okky-hi-fix-emphasize-editable-fields-r06-edge.vercel.app/apps)
2. Create a new app or open one you've already created.
3. Create a new Generator Node or open one you've already created.

## Other Information

It's common practice to place edit icons and action menus on the right side of editable information — for example, in GitHub's Pull Request title editing feature.
By placing the edit icon on the right, we aim to make it more intuitive for users to recognize that the content is editable by clicking the area.

Github
名称未設定

Vecel
名称未設定6


internal-packages/workflow-designer-ui/src/editor/properties-panel/ui/editable-text.tsx
```diff
@@ -68,6 +71,9 @@ export function EditableText({
 				onClick={() => setEdit(true)}
 			>
 				{value ?? fallbackValue}
+				{widthIcon && (
+					<Pen aria-label="Edit name" className="shrink-0 size-4" />
+				)}
```
okky-hi: Icons are conditionally rendered so they are not displayed when unnecessary.

名称未設定3

internal-packages/workflow-designer-ui/src/editor/properties-panel/ui/editable-text.tsx
```diff
@@ -1,16 +1,19 @@
...
 export function EditableText({
 	value,
 	fallbackValue,
 	onChange,
+	widthIcon,
 }: {
 	value?: string;
 	fallbackValue: string;
 	onChange?: (value?: string) => void;
+	widthIcon?: boolean;
```
toyamarinyon: `withIcon`?

toyamarinyon (commented):
I wasn't sure whether to prioritize the design aspect of having the icon always visible or to prioritize clarity. I'll leave the decision to @kaochannel154 .

kaochannel154:
@okky-hi 
Thank you for implementing it! I've checked it out.
There is definitely a slight sense of incongruity with the fact that it is always displayed, so it might be better to refer to other products.
I prefer the GitHub treatment, but can it be changed? I'm sorry that you came up with it🙏

okky-hi:
@kaochannel154 
Thanks for the feedback!
I have three questions.

1. When you say "I prefer the GitHub treatment," do you mean you'd like the UI styled similarly to GitHub's design?
2. For example, are you envisioning something like GitHub’s top navigation/search bar—as shown in the screenshot I shared—where icons are not always visible, but appear (or show a tooltip) on hover?
3. If my understanding is off, could you possibly share a screenshot of the UI you're referring to, if that's not too much trouble? That would really help me better align with your vision.



名称未設定6

kaochannel154:
@okky-hi 
I believe this is the screenshot that was shared!
スクリーンショット 2025-04-04 11 29 40

I thought the smaller "edit" button might feel more familiar since it's commonly used in blogs and editing interfaces. What do you think?

PR #651, opened 2025-04-04
Fix (workaround) Enable to properly unset next action in GitHub Integration Setting

satococoa:
## Summary
I've corrected the process that resets the nextAction appropriately based on the trigger value.

## Related Issue
- https://github.com/radix-ui/primitives/issues/1569
- https://github.com/shadcn-ui/ui/discussions/638

### DIFF
--- a/internal-packages/workflow-designer-ui/src/settings/github-integration/github-integration-setting-form.tsx
+++ b/internal-packages/workflow-designer-ui/src/settings/github-integration/github-integration-setting-form.tsx
@@ -111,6 +111,16 @@ function Installed({
 		WorkspaceGitHubIntegrationPayloadNodeMap[]
 	>(data?.payloadMaps || []);
 
+	// Select component doesn't support resetting the value to undefined
+	// So we need to use a workaround
+	// https://github.com/radix-ui/primitives/issues/1569
+	// https://github.com/shadcn-ui/ui/discussions/638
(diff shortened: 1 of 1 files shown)


toyamarinyon (commented):
This is a workaround, but what is the proper (or standard) way to do this, and what challenges/issues are making it a workaround?"

shige:
📝 I will do my OATs in preview branch.

shige (approved):
Thank you! 👍

satococoa:
@toyamarinyon 

> This is a workaround, but what is the proper (or standard) way to do this, and what challenges/issues are making it a workaround?

The proper way would be to simply set `""` or `undefined` to the `nextAction` value. 

However, when doing so, the form value of `nextAction` is not actually cleared. Instead, the first value in the SelectItem list is silently used as the form value, which is not the expected behavior.

This appears to be a specification issue with Radix UI's Select component. It's a known issue that has been documented in:
- https://github.com/radix-ui/primitives/issues/1569
- https://github.com/shadcn-ui/ui/discussions/638

Due to this limitation, our current implementation serves as a workaround until the underlying issue is addressed in the library itself.

toyamarinyon (approved):
I understand the approach now - thanks for the clear explanation about the Radix UI Select component limitations! This is a good workaround until the underlying library issue is resolved.

PR #688, opened 2025-04-10
fix(studio.giselles.ai): fallback to the first team if the specified team is not found

satococoa:
## Summary
Issue occurs when selecting a team from the team selector in the top-right corner.

Potential scenarios causing the error:
1. User is removed from the team membership between team list fetch and team selection
2. Team is deleted between team list fetch and team selection
3. User is removed due to team downgrade to free plan

The issue is related to race conditions between fetching the team list and selecting a team.


shige (approved):
Thank you! 👍

toyamarinyon (approved):

PR #704, opened 2025-04-14
GitHub tool

toyamarinyon:
**Important Note**: This is currently in the development team's verification phase. Detailed instructions and warnings regarding PAT (Personal Access Token) usage are not yet included. Crucially, PATs are currently being stored in plain text. This needs to be addressed before moving to production due to security concerns.


shige:
I would like to ask @satococoa san to review the code as well!

apps/playground/giselle-engine.ts
satococoa, line 1: Despite these changes, our GitHub integrations are still working properly.

- Users can set up GitHub Webhook triggers for their workspace
- GitHub Webhook triggers continue to be handled properly

Is my understanding correct?

packages/giselle-engine/src/http/router.ts
```diff
@@ -220,6 +220,13 @@ export const createJsonRouters = {
 				return JsonResponse.json(workspace);
 			},
 		}),
+	getGitHubRepositories: (giselleEngine: GiselleEngine) =>
```
satococoa, line 223: What's the purpose of this endpoint?(for debugging?)

toyamarinyon (commented):
@satococoa Just to clarify (as is evident from the code), the PAT used for the `Text Generation` node's GitHub Tool feature is entirely separate from the main GitHub Integration.

While rigorous interaction testing hasn't been performed, they were designed and implemented to be isolated. We expect no impact on the main GitHub Integration's stability, although a small theoretical risk of unforeseen side effects exists.

satococoa (approved):
Got it, thanks 👍

shige (approved):
Thank you! 

⚠️ This is an experimental PR. It does not affect regular users. ⚠️ 
The part where the PAT is stored in plain text will be improved for secure storage by the time of the public release.

PR #717, opened 2025-04-15
Remove Next.js config for logging and server actions

toyamarinyon:
## Summary
- Remove unneeded Next.js config for logging and server actions
- Replace require-in-the-middle hoist pattern with direct dependency

## Test plan
- Verify that the application builds and runs properly
- Confirm that there are no regressions in server-side functionality


apps/studio.giselles.ai/next.config.ts
```diff
@@ -6,16 +6,6 @@ const nextConfig: NextConfig = {
...
 		ignoreDuringBuilds: true,
 	},
-	logging: {
-		fetches: {
-			fullUrl: true,
-		},
-	},
-	experimental: {
-		serverActions: {
-			bodySizeLimit: "4.5mb",
```
shige, line 16: Question! 👂 @toyamarinyon 

According to https://nextjs.org/docs/app/api-reference/config/next-config-js/serverActions#bodysizelimit:

> By default, the maximum size of the request body sent to a Server Action is 1MB,

Is it okay to remove this statement?

toyamarinyon: This setting appears to be a remnant from when Server Actions were used for file uploads. We've since migrated to Route Handlers, so it's likely no longer necessary. However, I'll verify this in the preview environment to confirm.

toyamarinyon: @shige I've checked uploading 4.1MB PDF and generate with this file at https://giselle-git-improve-dev-r06-edge.vercel.app/ !

shige: I see, I understand!

---

📝 Related codes:

https://github.com/giselles-ai/giselle/blob/481a0c1aa34811999a68ecc89a875dd5f46c75f0/internal-packages/workflow-designer-ui/src/editor/properties-panel/file-node-properties-panel/file-panel.tsx#L22-L27

shige (approved):
Thank you! 🚿

PR #728, opened 2025-04-17
Fix: Skip non-uploaded files in copyFiles function

toyamarinyon:
## Summary
- Add condition to skip files that don't have 'uploaded' status in the copyFiles function
- Prevents errors when trying to copy non-existent files during run creation

## Test plan
- Create a workflow with file nodes having different status values
- Verify that only files with 'uploaded' status are copied to the run
- Check that no errors occur when processing file nodes with non-uploaded files


shige (approved):
Thank you! 🚀

PR #732, opened 2025-04-17
Fix: Prevent duplicate file attachments in message generation

toyamarinyon:
## Summary
- Prevent duplicate file attachments when the same file is referenced multiple times in a prompt
- Add tracking of attached file node IDs in the text generation message builder
- Only attach each file once to avoid potential issues with LLM providers

## Test plan
- Create a prompt that references the same file multiple times
- Verify that the file is only attached once in the generated message
- Check that the placeholder text for the file is still correctly inserted in each reference


shige (approved):
Thank you! 👍

PR #738, opened 2025-04-18
Add feature flag to control share UI components

toyamarinyon:
Make UserPresence and ShareButton rendering conditional based on a new shareFeatureFlag prop. Added a TODO note to replace this with a proper feature flag provider in the future.


shige (approved):
Thank you! 🎏

PR #744, opened 2025-04-22
Update OpenAI models to use OptionalSearchGrounding capability

toyamarinyon:
## Summary
- Changed several OpenAI model definitions to use OptionalSearchGrounding instead of SearchGrounding capability flag
- This change indicates that web search is optional for these models rather than mandatory

## Test plan
- Verify that GPT-4o, GPT-4o-mini, GPT-4.1, and GPT-4.1-mini models correctly show web search as an optional capability
- Ensure the web search toggle functionality works correctly with these models


shige:
@toyamarinyon I don't understand the intention of the change, so I have a question.

Is the intended fix to display the message `Web search will not use since the current model does not support web search`?

I thought that when this message was displayed, users would be confused because they would not know whether they could use "Web Search" or not.

| Preview | Production |
|--------|--------|
| image | image |

toyamarinyon:
@shige Fixed!

shige (approved):
Thank you! 💯

PR #749, opened 2025-04-23
feat: add auto-dismiss behavior to toasts after 5 seconds

gentamura:
Implementing global event handling to dismiss error messages on click events would require complex state management across the application. To avoid this, we opted for a simpler solution of auto-dismissing toast messages after 5 seconds.

This change:
- Adds automatic dismissal of toast messages after 5 seconds
- Simplifies error state management
- Maintains application architecture without introducing global event handlers

### DIFF
--- a/internal-packages/workflow-designer-ui/src/ui/toast.tsx
+++ b/internal-packages/workflow-designer-ui/src/ui/toast.tsx
@@ -43,7 +43,7 @@ export const ToastProvider: React.FC<{ children: React.ReactNode }> = ({
 
 	const error = useCallback(
 		(message: string) => {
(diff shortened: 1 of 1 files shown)


gentamura:
@shige @toyamarinyon review, please :pray:

toyamarinyon (commented):
Toast notifications should not close automatically. If a user switches to another tab while an operation is running and returns later, they might miss error messages or other important status updates if the toast has already disappeared. This makes it difficult to determine if an error occurred and what the details were.

gentamura:
Based on the above comments, I'm closing this PR.

PR #755, opened 2025-04-23
Added Trigger Node to Toolbar

toyamarinyon:
I've added a Trigger Node to the toolbar, controlled by a Feature Flag. By default, it's hidden and only visible to the development team.

The node is always visible in dev mode on Playground and studio.giselles.ai.

Note: This PR only implements the node addition to the UI. The execution control functionality will be implemented separately.


shige (approved):
Thank you! 🎨

PR #757, opened 2025-04-23
Added Feature Flag `RunV2` for New Run Button Implementation

toyamarinyon:
I've added a new Feature Flag called `RunV2` that allows us to toggle between the current header display and a new implementation for the Run button. This PR sets up the infrastructure for the new Run button functionality, though the actual implementation of the new button is not included in this PR and will be developed separately.

Enable RUN_V2|Disable Run V2
---------------|----------------
image|image

This change enables us to:
- Safely develop the new Run button behind a feature flag
- Test both implementations in parallel
- Gradually roll out the new feature when ready

The current functionality remains unchanged when the feature flag is disabled.


shige (approved):
Thank you! 👍

PR #759, opened 2025-04-24
Add Claude Code GitHub workflow

toyamarinyon:


shige (approved):
👍 

Since the API key is stored in secrets, I want to delete it when it's no longer needed.

PR #764, opened 2025-04-24
Added Reasoning option for Claude 3.7

toyamarinyon:
This PR adds support for the Reasoning option (called "Thinking" by Anthropic) when using Claude 3.7. The property name has been set as "Reasoning" to maintain consistency with the AI SDK naming convention.

Note that this option is silently ignored when using Claude models that don't support this feature.


shige (approved):
Thank you! 🧠

PR #766, opened 2025-04-24
Added Manual Trigger Node Type

toyamarinyon:
I've added a new "Manual" type to the Trigger node. While it doesn't have any functionality yet, I'm merging this now because it's important to establish that we have trigger node types beyond just GitHub. Further functionality will be implemented in future updates.


shige (approved):
Thank you! 🌃

PR #769, opened 2025-04-25
# Added support for gpt-image-1 model

toyamarinyon:
Added support for gpt-image-1 model

I've implemented support for the gpt-image-1 model. Previously, our Image Generation feature was exclusively using Fal, but with the addition of OpenAI as a provider, I've restructured the code to branch processing based on the selected provider.

While we're tracking costs for Fal through Langfuse, we haven't yet set up cost calculation for gpt-image-1. Therefore, I've implemented this behind a feature flag to prevent full public release. You can test the functionality by enabling the `gpt-image1` feature flag.


shige:
@toyamarinyon 

> While we're tracking costs for Fal through Langfuse, we haven't yet set up cost calculation for gpt-image-1. Therefore, I've implemented this behind a feature flag to prevent full public release. You can test the functionality by enabling the `gpt-image1` feature flag.

It's okay to proceed with the public release without a flag. 
Our understanding is that this isn't an issue because we are only providing it under the Pro plan!

shige (approved):
Thank you! 🖼️

PR #773, opened 2025-04-26
chore(biome): refactor biome

hiroppy:
## Summary

- refering schema in node_modules to easy to match version
- some biome.json in packages extend root biome.json to remove same rules
- adding biome to playground since CI checks this package

### DIFF
--- a/apps/studio.giselles.ai/biome.json
+++ b/apps/studio.giselles.ai/biome.json
@@ -1,18 +1,6 @@
 {
-	"$schema": "https://biomejs.dev/schemas/1.8.2/schema.json",
-	"organizeImports": {
-		"enabled": true
-	},
-	"linter": {
-		"enabled": true,
-		"rules": {
-			"recommended": true
-		}
-	},
-	"vcs": {
-		"enabled": true,
-		"clientKind": "git"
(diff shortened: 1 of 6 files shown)


apps/studio.giselles.ai/biome.json
```diff
@@ -1,18 +1,6 @@
 {
-	"$schema": "https://biomejs.dev/schemas/1.8.2/schema.json",
```
hiroppy, line 2: the current version is 1.9.4 in studio's package.json so it's mismatched. we should refer node_modules, not external URL

apps/studio.giselles.ai/package.json
```diff
@@ -10,7 +10,7 @@
 		"dev-debug": "LOGLEVEL=debug pnpm run dev",
 		"build": "next build --turbopack",
 		"start": "next start",
-		"check": "biome check --write .",
```
hiroppy, line 13: studio has "check" script but all packages have "format" and the root turbo command is also "format"

toyamarinyon (approved):

shige (approved):
Thank you! 🚀

PR #776, opened 2025-04-27
feat(studio): add some security headers

hiroppy:
## Summary

ssia, see https://nextjs.org/docs/pages/api-reference/config/next-config-js/headers#options


toyamarinyon (approved):

shige (approved):
Thank you! 🚀

PR #777, opened 2025-04-27
fix(studio): rename check to format

hiroppy:
## Summary

studio has "check" script but all packages have "format" and the root turbo command is also "format"


hiroppy:
closed by https://github.com/giselles-ai/giselle/pull/777

PR #779, opened 2025-04-27
refactor(studio, playground): use * for env

hiroppy:
## Summary

this repo has name rules for env variables so we can use * to prevent forgetting to add new variable.


toyamarinyon (approved):
I didn't know that `*` could be used! Thank you!

shige (approved):
Thank you! 🚀

PR #780, opened 2025-04-28
Action to operation

toyamarinyon:
Node Data Structure Update: "action" to "operation"

We have changed the hierarchical structure of Node data from "action" to "operation".

### Testing Requirements
To verify that the data modification process is functioning correctly, please:

- Confirm that existing Apps can be displayed without errors
- Verify that Generation can be executed successfully


shige (approved):
Thank you! 🚀 

I have confirmed the following to work as a minimum! ✅ 

- Existing giselle apps work
- Newly created giselle apps work

Furthermore, we would like to make it a policy to fix any errors as soon as they are notified.

PR #781, opened 2025-04-28
Action Node

toyamarinyon:
Added an Action node to the system.

> [!note]
> This is only displayed when the flow flag is enabled in feature flags. It is not visible to general users.

https://github.com/user-attachments/assets/1f4a816d-ebde-40fe-835a-0c1061b7e97d


shige (approved):
Thank you!

PR #782, opened 2025-04-29
Generation state on node

toyamarinyon:
https://github.com/user-attachments/assets/57700a2e-2cf4-4990-a57a-8d770ad3056e


shige (approved):
Thank you! 🔁

PR #783, opened 2025-04-30
Add helper function for the flow

toyamarinyon:
I have implemented a function that retrieves connected nodes starting from any node and constructs a flow. This will be necessary for Triggers.


shige (approved):
Thank you! 🧰

PR #785, opened 2025-04-30
Execute flow with Manual Trigger as the starting point

toyamarinyon:
We have enabled the flow to be executed with Manual Trigger as the starting point. This feature requires a feature flag and will not be available to general users.

https://github.com/user-attachments/assets/c998c8bb-04d8-48f6-90d3-ff3157b846d3


shige (approved):
Thank you! 🏃

PR #789, opened 2025-04-30
Refactor node generation fetching with SWR

toyamarinyon:
## Summary
- Added SWR dependency for efficient data fetching with caching and revalidation
- Refactored node generation fetching from context to hook implementation
- Improved state management by exposing setGenerations in context

## Test plan
- Run the application and verify that node generations load correctly
- Check that generation data updates properly when status changes
- Verify that generation data persists between component re-renders

### DIFF
--- a/packages/giselle-engine/package.json
+++ b/packages/giselle-engine/package.json
@@ -72,6 +72,7 @@
 		"openai": "catalog:",
 		"pg": "catalog:",
 		"react": "catalog:",
+		"swr": "catalog:",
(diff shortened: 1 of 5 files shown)


toyamarinyon:
I would like to add background context for why I created this pull request.

We were executing an API that only needs to be loaded once every time the node state changed, so I made it cacheable. In the future, I think it would be even better if it could be loaded during server-side rendering.

shige (approved):
Thank you! 📝

PR #794, opened 2025-05-01
Add connector animation for running input nodes

toyamarinyon:
## Summary
- Add visual feedback to connectors by showing an animated gradient when the input node is in 'queued' or 'running' state
- Implement conditional rendering based on node generation status using existing hooks from the SDK
- Improve user experience by making the workflow process state more visible

https://github.com/user-attachments/assets/2f5fe23e-d3f3-460e-9a2c-a89ac07ad33c

## Test plan
- Run the workflow designer and verify that connectors show animation when their input node is in running or queued state
- Confirm the animation disappears when the node is no longer processing
- Test with multiple connected nodes to ensure correct behavior across the workflow


shige (approved):
Thank you! 🌈

PR #798, opened 2025-05-01
Add reasoning styles

kaochannel154:
## Summary
Improve UI styling for the Thinking Process display by removing unnecessary visual elements for better readability and cleaner appearance.
スクリーンショット 2025-05-01 20 56 23

スクリーンショット 2025-05-01 20 57 16

スクリーンショット 2025-05-01 20 57 24

## Related Issue
N/A

## Changes
- Improved styling for the Reasoning switch in the property panel
- Added alert for models that don't support reasoning
- Implemented progressive styling for Thinking Process display:

## Testing
- Tested the UI changes by confirming that:
  - Active reasoning still shows the Willi icon and curly brace with "Thinking..." text
  - Completed reasoning shows only "Thinking Process" text without icon or brace
  - Alert appears properly for non-supported models
  - Transitions between states work correctly
  - Property panel switch styling displays correctly

## Other Information
This change is part of ongoing UI improvements to make the reasoning feature more intuitive and visually appealing.

### DIFF
--- a/internal-packages/workflow-designer-ui/src/editor/properties-panel/text-generation-node-properties-panel/model/anthropic.tsx
+++ b/internal-packages/workflow-designer-ui/src/editor/properties-panel/text-generation-node-properties-panel/model/anthropic.tsx
@@ -5,7 +5,7 @@ import {
 	hasCapability,
 } from "@giselle-sdk/language-model";
 import { useUsageLimits } from "giselle-sdk/react";
-import { useMemo, useState, useEffect, useRef } from "react";
+import { useEffect, useMemo, useRef, useState } from "react";
 import {
 	Select,
 	SelectContent,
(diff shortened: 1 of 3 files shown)


toyamarinyon:
Great!!!!

shige (approved):
Thank you!! 💭

toyamarinyon:
Sorry. I'll review later. We should think when and how to use ghost icon

toyamarinyon (commented):
I think there are two things we should consider.

1. It feels awkward when discrete animations and continuous animations are mixed within a single sequence.
   
   After pressing the Generate button, the animation where three Ghosts appear in a row moves discretely with a blinking effect. This is very cute.
   
   Next, when Thinking begins, it displays with a shimmer effect. It's very beautiful. The shimmer representation of thinking is commonly seen, so it feels natural.
   
   I think both animations are well made on their own, but when combined, they seem a bit disjointed.
   
   It's like starting with R&B and suddenly switching to a ballad. While you could say that's what aespa does, it would be surprising if aespa appeared here. That kind of thing stands out because it's on the center stage. I think in these cases, not changing the rhythm of animations too much creates a more natural and comfortable experience.
   
   Similarly with the Cursor, the `...` is a discrete animation and Thinking is a shimmer, but I didn't feel any awkwardness because they appear in different positions. Perhaps we feel awkwardness when animations displayed in the same position change rhythm.

    https://github.com/user-attachments/assets/32fc5da5-d47d-435b-a9de-9802d13ce2f3

2. It might be time to give meaning to the characters
   
   Until now, we've been using the ghost characters as a sort of design accent without much thought. We haven't really delved into it, but it might be time to consider it. This is because when the ghost appears during Thinking, I had the impression that it was doing the thinking. I don't know if it's one of the LLMs or something that manages them, but I think we need some kind of setting for this.
   
   Personally, I think we can express it more freely if it's a separate entity from the functionality.

internal-packages/workflow-designer-ui/src/ui/generation-view.tsx
```diff
@@ -72,9 +72,10 @@ export function GenerationView({
 				<div key={message.id}>
 					{message.parts?.map((part, index) => {
 						const lastPart = message.parts?.length === index + 1;
+						const isRunning = generation.status === "running";
 						switch (part.type) {
 							case "reasoning":
-								if (lastPart) {
+								if (lastPart && isRunning) {
```
toyamarinyon: Why do we need consider `isRunning` ?

internal-packages/workflow-designer-ui/src/editor/properties-panel/text-generation-node-properties-panel/model/anthropic.tsx
toyamarinyon, line 1: This implementation is too complicated for what I want to do, so I'll create a pull request to rewrite it a bit.

toyamarinyon (approved):

PR #804, opened 2025-05-07
Upgrade Flags SDK to v4

shige:
## Summary
I upgraded Flags SDK to v4.

## Related Issue
None.

## Changes
- Upgrade Flags SDK to v4
- Adjust flags discovery endpoint

## Testing
1. Visit https://giselle-git-upgrade-flags-sdk-to-v4-r06-edge.vercel.app/
2. Enable/Disable flags

## Other Information
- [Upgrade Flags SDK to v4](https://github.com/vercel/flags/blob/flags%404.0.0/packages/flags/guides/upgrade-to-v4.md)

I cannot use flags in production env.

image

### DIFF
--- a/apps/studio.giselles.ai/app/.well-known/vercel/flags/route.ts
+++ b/apps/studio.giselles.ai/app/.well-known/vercel/flags/route.ts
@@ -1,4 +1,4 @@
-import { getProviderData, createFlagsDiscoveryEndpoint } from "flags/next";
+import { createFlagsDiscoveryEndpoint, getProviderData } from "flags/next";
 import * as flags from "../../../../flags";
 
 export const runtime = "edge";
--- a/apps/studio.giselles.ai/app/.well-known/vercel/flags/route.ts
+++ b/apps/studio.giselles.ai/app/.well-known/vercel/flags/route.ts
@@ -1,14 +1,14 @@
-import { type ApiData, verifyAccess } from "flags";
-import { getProviderData } from "flags/next";
-import { type NextRequest, NextResponse } from "next/server";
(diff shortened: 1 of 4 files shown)


shige:
@toyamarinyon @satococoa 
Currently, I am unable to switch flags in the production environment, which is causing me trouble with the operation of my Giselle.
I would appreciate it if you could prioritize this in your review.

apps/studio.giselles.ai/app/.well-known/vercel/flags/route.ts
```diff
@@ -1,14 +1,14 @@
-import { type ApiData, verifyAccess } from "flags";
-import { getProviderData } from "flags/next";
-import { type NextRequest, NextResponse } from "next/server";
+import { createFlagsDiscoveryEndpoint, getProviderData } from "flags/next";
 import * as flags from "../../../../flags";
 
 export const runtime = "edge";
 export const dynamic = "force-dynamic"; // defaults to auto
```
satococoa, line 5: I don't know why these settings are here.
Do we need these settings still?

The functionality works fine in my testing, by the way. ✅

shige:
Thank you! 🚀

apps/studio.giselles.ai/app/.well-known/vercel/flags/route.ts
shige: @satococoa 

Thank you for your comment!
I understand that this definition is necessary because the flags are customized.

https://github.com/giselles-ai/giselle/blob/e07de6c778c1fb21d46813e74555601819e8d102/apps/studio.giselles.ai/app/.well-known/vercel/flags/route.ts#L2

I think this definition is unnecessary if you use flags normally.

PR #806, opened 2025-05-07
Remove unnecessary useState and useCallback hooks from UI components

toyamarinyon:
This PR includes two main improvements:

1. Removed excessive useState and useCallback hooks from UI components as they were unnecessary for this implementation. This change simplifies the code while maintaining the same functionality.

2. Added confirmation dialogs for file deletion with the following behavior:
   - Failed uploads: Delete immediately (no confirmation)
   - Uploading files: Show confirmation dialog before deletion
   - Uploaded files: Show confirmation dialog before deletion

These changes improve both code quality and user experience by simplifying the implementation and preventing accidental file deletions.


shige (approved):
Thank you! 👍

PR #743, opened 2025-04-21
Add OpenAI Web Search Tool integration for text generation

toyamarinyon:
## Summary
- Implements web search grounding for OpenAI models using the official `openai.tools.webSearchPreview()` API
- Adds a user-friendly "Web Search" toggle in the OpenAI model configuration panel
- Creates a new `OpenAIWebSearchTool` schema to define search parameters like context size and user location
- Introduces `GenerationContextLike` type for more flexible validation in generation workflows
- Updates OpenAI call implementation to use the modern `responses` namespace

I've checked works properly on https://giselle-playground-git-fork-toyamarinyon-openai-0b3436-r06-edge.vercel.app/

https://github.com/user-attachments/assets/d6242384-3b42-4298-bbbd-6807f4b7521d

## Test plan
1. Enable the "Search Grounding" toggle in the OpenAI model panel
2. Create a text generation that would benefit from real-time web data (e.g., "What are the latest news about AI regulation?")
3. Verify that the generated text includes recent information with proper citations
4. Test with different search context sizes (low, medium, high) to observe differences in result quality
5. Ensure the feature works correctly in both playground and studio environments

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #809, opened 2025-05-07
Tag Langfuse trace with web-search info

Rindrics:
## Summary

Tag Langfuse trace if user executes OpenAI node with "Web Search" enabled

## Related Issue

- https://github.com/giselles-ai/giselle/pull/743

## Testing

The trace is tagged with `web-search` if the functionality is used ✅ 
image

### DIFF
--- a/packages/giselle-engine/src/core/generations/generate-text.ts
+++ b/packages/giselle-engine/src/core/generations/generate-text.ts
@@ -482,7 +482,12 @@ export async function generateText(args: {
 		},
 		experimental_telemetry: {
 			isEnabled: args.context.telemetry?.isEnabled,
-			metadata: args.telemetry?.metadata,
+			metadata: {
+				...args.telemetry?.metadata,
+				...(preparedToolSet.toolSet.openaiWebSearch
+					? { tags: ["web-search"] }
+					: {}),
+			},
 		},
 	});
 	return streamTextResult;


packages/giselle-engine/src/core/generations/generate-text.ts
```diff
@@ -482,7 +482,12 @@ export async function generateText(args: {
 		},
 		experimental_telemetry: {
 			isEnabled: args.context.telemetry?.isEnabled,
-			metadata: args.telemetry?.metadata,
+			metadata: {
+				...args.telemetry?.metadata,
+				...(preparedToolSet.toolSet.openaiWebSearch
+					? { tags: ["web-search"] }
+					: {}),
+			},
```
toyamarinyon, line 490: It looks great and seems adaptable for use with other tools as well.

shige (approved):
Thank you! 🏷️ 

---

ℹ️ FYI: 

It seems that a property called OpenaiWebSearch is automatically generated in Langfuse Generation.
As Giselle's o11y, I think it's fine to leave the policy of explicitly tagging `web-search` across models as is.


```json
{
    "type": "provider-defined",
    "id": "openai.web_search_preview",
    "args": {
        "searchContextSize": "medium"
    }
}
```

image

packages/giselle-engine/src/core/generations/generate-text.ts
rindrics: Thank you, you're right.

I'll try to uniform them in another PR!

rindrics:
thank you for your reviews!

PR #811, opened 2025-05-08
Remove a bit odd text

shige:
## Summary
I remove a bit odd text.

## Related Issue
None.

## Changes
I remove the text `​Thank you for being`.

| main branch | this pr |
|--------|--------|
| image | image | 

## Testing
1. Vist /settings/team page

### Feedback by proofreader

Thank you for being Pro Plan seems a bit odd. I think Thank you for being Pro would be more natural, but the style is a bit different from other menus and titles, so it seems like it would be fine without it, just like Free.


shige:
@Attolis Could you please review it when you have time?

toyamarinyon (approved):

shige:
Thank you! 🚀

PR #813, opened 2025-05-08
Fix: Handle undefined text generation output consistently

toyamarinyon:
## Summary
- Update text generation output handling to replace references with empty strings when output is undefined
- Add comment explaining the behavior for better maintainability
- Ensure consistent behavior across different node types

## Test plan
- Verify that text generation nodes with undefined outputs replace the reference pattern with an empty string
- Ensure this behavior is consistent with other node types in the system


shige (approved):
Thank you! 🆙

PR #814, opened 2025-05-08
Refactor Anthropic model panel reasoning capability UI for improved extensibility

toyamarinyon:
This PR addresses issues with the reasoning component implementation in the Anthropic model panel:

## Changes
- Refactored the reasoning component to eliminate the use of useEffect and complex inline callbacks
- Improved the architecture for better future extensibility and performance
- Made subtle enhancements to the UI appearance for better user experience
- Removed generation running check for reasoning part

## Reasoning
The previous implementation was tightly coupled and relied heavily on side effects, which could become difficult to maintain and extend as new features are added. This refactoring creates a more maintainable approach while preserving all existing functionality.

## Visual Changes

https://github.com/user-attachments/assets/1a90c9c8-f8fc-4b18-8b15-7384464914c7


shige (approved):
Thank you! 🆙

kaochannel154 (approved):
I think it's good, thank you👻！！

PR #822, opened 2025-05-09
proofreading: OpenAI web search unsupported error message

Attolis:
## Summary

Rewords the error message displayed when the OpenAI model doesn't support web search.

## Related Issue

None

## Testing

- Open any App and add an OpenAI generation node
- Go to the `Model` tab and turn on `Web Search`
- Change the model to `o3` and confirm the error message

image


Attolis:
I updated the message shown on OpenAI generation nodes when the model doesn't support web search. Could you please review this when you have a chance? @shige

shige (approved):
Thank you! 🚀

Issue #823, opened 2025-05-09
Send Langfuse trace from `gpt-image-1` node


Rindrics:
## Task description

Send Langfuse trace from `gpt-image-1` node

## Background

- Image generation via `gpt-image-1` is introduced by https://github.com/giselles-ai/giselle/pull/769
  - But we cannot track its usage data because the node does not emit Langfuse trace

rindrics:
- done by #824

PR #824, opened 2025-05-09
Send Langfuse trace from `gpt-image-1` node to track its usage and cost

Rindrics:
## Summary

Enables tracking usage and cost of `gpt-image-1` node

## Related Issue

- #823

## Changes

- Instrument Langfuse tracer to `generateImageWithOpenAI()` to send trace from `gpt-image-1` node
- Use fal.ai API via AI SDK to minimize dependency difference between providers

## Testing

Corresponding trace with calculated cost arrived at Langfuse ✅ 
image

## Other Information

- I didn't introduce a generic function to unify provider-specific function calls to keep the code diff in this PR small

### DIFF
--- a/packages/language-model/src/openai-image.ts
+++ b/packages/language-model/src/openai-image.ts
@@ -1,5 +1,5 @@
 import { z } from "zod";
-import { Capability, LanguageModelBase } from "./base";
+import { Capability, LanguageModelBase, type UsageCalculator } from "./base";
 
 export const size = z.enum(["1024x1024", "1536x1024", "1024x1536"]);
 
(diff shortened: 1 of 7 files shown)


packages/language-model/src/openai-image.ts
```diff
@@ -46,3 +46,59 @@ export const models = [openaiGptImage1];
...
+		"1536x1024": 0.063,
+	},
+	high: {
+		"1024x1024": 0.167,
+		"1024x1536": 0.25,
+		"1536x1024": 0.25,
+	},
+	auto: {
+		// price is set same as "medium" because we cannot detect which quality is finaly adopted by OpenAI
+		// ref: https://platform.openai.com/docs/guides/image-generation?image-generation-model=gpt-image-1#customize-image-output
```
rindrics, line 78: set price of `auto` same with that of `medium`'s

because (current version of) `generateImage` of AISDK does not returns which quality is selected by OpenAI
ref: https://ai-sdk.dev/docs/reference/ai-sdk-core/generate-image#responses

toyamarinyon (approved):

packages/language-model/src/openai-image.ts
```diff
@@ -46,3 +46,59 @@ export const models = [openaiGptImage1];
...
 export const LanguageModel = OpenAIImageLanguageModel;
 export type LanguageModel = OpenAIImageLanguageModel;
+
+export interface OpenAIImageGenerationParams {
+	width: number;
+	height: number;
+	quality: "auto" | "low" | "medium" | "high";
+}
+
+const openAICostTable: Record<
```
shige, line 56: @Rindrics We would appreciate it if you could include the URL of the document from which the amounts in openAICostTable are calculated.

shige (approved):
Thank you! 📊

packages/language-model/src/openai-image.ts
rindrics: fixed: https://github.com/giselles-ai/giselle/pull/824/commits/2c0a38c646c19fdd1d6d3903ebd9676d9e1bbbfe

rindrics:
thank you for your reviews!

PR #825, opened 2025-05-09
Optimize useSWR key structure in useGitHubTrigger hook

toyamarinyon:
## Summary
- Simplify useSWR key structure for better readability and maintainability 
- Refactor first useSWR call to use string key instead of array format
- Refactor second useSWR to use object-based key instead of array format
- Remove unnecessary nested lambda function in first useSWR call

## Test plan
- Verify GitHub repositories properly load in the workflow designer
- Ensure trigger configurations continue to function correctly
- Confirm proper data fetching behavior is maintained


shige (approved):
Thank you! ⚡

PR #826, opened 2025-05-10
Filter out unconfigured trigger nodes from the header

toyamarinyon:
## Description
This PR filters out unconfigured trigger nodes from being displayed in the workflow header. Only trigger nodes that have been properly configured will now appear in the header section.

## Implementation Details
- Modified the trigger nodes filtering logic in the `Trigger` component
- Added a check to exclude nodes with `status === "unconfigured"`
- Improved code readability by using a more explicit loop instead of a filter function

## Test Plan
1. Create a workflow with a trigger node
2. Verify that unconfigured trigger nodes are not displayed in the header
3. Configure the trigger node and verify it appears in the header

## Additional Notes
This change improves the user experience by only showing actionable triggers in the header.


shige (approved):
Thank you! ⚙️

PR #812, opened 2025-05-08
Add GitHub Vector Store (Blobs only)

satococoa:
> [!note]
> migrations have been already applied in #832 

## Summary
Added functionality to import GitHub repository code into Vector Store.
This allows teams to use GitHub code as a searchable Vector Store.

The management screen enables repository registration and deletion, and we've implemented a mechanism for periodic data ingestion using Vercel Cron.

Because the Vector Store feature is gated behind a feature flag, it is hidden from regular users.

> [!tip]
> The initial ingestion of the *giselle* repository took **11 minutes**.

### studio.giselles.ai
* Added `/settings/team/vector-stores` page
   * Implemented Vector Store registration and deletion functionality
* Added `/api/vector-stores/github/ingest` API endpoint
   * Configured Vercel Cron to run automatically every 10 minutes
   * Note: Vercel Cron does not work in Preview environments, so manual execution is required
   * Manual execution example: `curl http://localhost:3000/api/vector-stores/github/ingest -H 'authorization: Bearer <CRON_SECRET>'`

### packages/rag
* `ingest()` orchestrator: load → chunk → embed -> store

### `packages/github-tool`
* Added `GitHubBlobLoader`

## Future Work
* Implement Vector Store Node in Workspace
* Implement differential ingestion
* Introduce job queue system

## Screenshots
スクリーンショット 2025-05-13 10 39 13

スクリーンショット 2025-05-13 10 39 19

スクリーンショット 2025-05-13 10 39 27

スクリーンショット 2025-05-13 10 39 42

PR #832, opened 2025-05-12
Add GitHub repository index and embeddings tables to schema

satococoa:
## Summary
This PR extracts the table design and migration from the implementation used in #812.
It adds the schema necessary for indexing GitHub repository code and enabling vector search.

### Added Tables

1. `githubRepositoryIndex` - Table for managing repository index information
2. `githubRepositoryEmbeddings` - Table for storing repository code embeddings

### Embedding Design

In the `githubRepositoryEmbeddings` table, the vector dimension is set to 1536.
This value matches the output dimension of OpenAI's embedding model `text-embedding-3-small`.

### Index Design

To improve vector search performance, we've added an index using the HNSW (Hierarchical Navigable Small World) algorithm:

```sql
embeddingIndex: index().using(
  "hnsw",
  table.embedding.op("vector_cosine_ops"),
),
```

The reason for using cosine similarity is that it's recommended by OpenAI in their official documentation:

https://platform.openai.com/docs/guides/embeddings#which-distance-function-should-i-use

> ## Which distance function should I use?
> We recommend cosine similarity.

After comparing the two major vector index methods, HNSW and IVFFlat, we chose HNSW for operational reasons:

- **No initial data required**: Unlike IVFFlat, HNSW indexes can be created even when the table is empty, making initial setup easier
- **Fewer parameter adjustments**: IVFFlat requires continuous adjustment of appropriate list numbers and probe counts, while HNSW performs well with default parameters
- **Adaptability to data changes**: Even when data volume or distribution changes, HNSW requires less frequent index rebuilding and maintains consistent search quality

Additionally, according to Neon's documentation, the IVFFlat method "requires creating the index after the table has some data," making it impractical for our use case where we need to build the index in the initial state.

For these reasons, we've adopted the HNSW index, which has less configuration and operational burden and is more resilient to data changes.

References:
* https://orm.drizzle.team/docs/guides/vector-similarity-search
* https://platform.openai.com/docs/guides/embeddings
* https://neon.tech/docs/extensions/pgvector

## TODO

 - [x] Run migration on Preview
 - [ ] Run migration on Production


toyamarinyon (approved):

shige (approved):
Thank you! 🔻

satococoa:
Thank you for your reviews.

@shige 
Could you merge this and run migration on Production, please?

shige:
> Could you merge this and run migration on Production, please?

@satococoa I did it! 🚀

Issue #796, opened 2025-05-01
Calculate model "cost" (not usage) inside app


Rindrics:
## Task description

We have to calculate model cost in dollar inside Giselle to tracking model costs correctly

## Background

- Currently, we do not calculate model costs ourselves, as Langfuse handles cost estimation.
  - These costs are synced to our data warehouse via a batch process
- For models whose pricing is not defined in Langfuse, we can calculate costs using a [custom model pricing function](https://langfuse.com/changelog/2024-01-29-custom-model-prices)
- However, relying on custom model definitions is not a long-term solution, as it cannot support complex pricing structures like [Gemini's multi-tier pricing](https://ai.google.dev/gemini-api/docs/pricing)
  - Furthermore, custom-defined pricing requires us to update prices synchronously whenever the model provider revises their pricing, which adds operational overhead

To resolve these problems, we want to:
- manage model price with `validateFrom` field inside our app
- calculate model cost in USD and send it to Langfuse

shige:
The core cost calculation infrastructure has been implemented for the main models:

**Completed:**
- OpenAI models (PR #834)
- Anthropic models (PR #872)
- Google models (PR #916)

**Remaining:**
- Perplexity models
- fal.ai models
- gpt-image

Since the primary models are covered, closing this issue for now. The remaining models can be tracked in a separate issue if needed.

PR #834, opened 2025-05-12
Calculate cost (for display, not for billing) for OpenAI text-generation models

Rindrics:
## Summary

enables:
- cost calculation of OpenAI text-generation models inside Giselle
  - for displaying preliminary value, not for billing
- telemetry emission for  Langfuse

## Related Issue

- part of #796

## Changes

- add pricing table for gpt models
- implement token-based calculation logic for OpenAI text-generation models
- introduce explicit Langfuse trace

## Testing

- corresponding trace arrived at Langfuse ✅ 
  - image
  - auto-instrumented trace from AISDK also arrives
    - we can distinguish the auto-instrumented trace with tag:
      - image

    - we will deprecate the auto-instrumented trace after #796 is completed

## Other information

We will implement cost calculation for billing in https://github.com/giselles-ai/giselle/issues/865


packages/language-model/src/costs/calculator.ts
```diff
@@ -0,0 +1,34 @@
...
+ * and be defined in their respective provider files.
+ */
+export interface CostCalculator<TUsage extends ModelUsage = TokenUsage> {
+	calculate(model: string, usage: TUsage): Promise<CostResult>;
+}
+
+export class DefaultCostCalculator implements CostCalculator {
+	constructor(private provider: string) {}
+
+	async calculate(): Promise<CostResult> {
```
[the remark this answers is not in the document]
rindrics: fixed: ed9bc828

packages/language-model/src/costs/pricing.ts
```diff
@@ -0,0 +1,63 @@
...
+export type TokenBasedPricing = {
+	input: BaseTokenPrice;
+	output: BaseTokenPrice;
+};
+
+export type ModelPrice = {
+	validFrom: string;
+	price: TokenBasedPricing;
+};
+
```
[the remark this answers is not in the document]
rindrics: fixed: e76f5953

packages/language-model/src/index.ts
```diff
@@ -25,6 +25,9 @@ import {
 } from "./perplexity";
 export * from "./base";
 export * from "./helper";
+export * from "./costs";
+export * from "./costs/pricing";
+export * from "./costs/model-prices";
```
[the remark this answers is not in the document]
rindrics: fixed: 81390a5f

packages/giselle-engine/src/core/generations/generate-text.ts
```diff
@@ -59,6 +60,9 @@ export async function generateText(args: {
 	if (!isTextGenerationNode(operationNode)) {
 		throw new Error("Invalid generation type");
 	}
+
+	const langfuse = new Langfuse();
```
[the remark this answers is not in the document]
rindrics: fixed: edfa68a1

packages/language-model/src/costs/model-prices.ts
```diff
@@ -0,0 +1,136 @@
...
+	modelId: string,
+	priceTable: ModelPriceTable,
+): ModelPrice {
+	const modelPricing = priceTable[modelId];
+	if (!modelPricing) {
+		throw new Error(`No pricing found for model ${modelId}`);
+	}
+
+	const now = new Date();
+	const validPrices = modelPricing.prices
```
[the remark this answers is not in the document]
rindrics: fixed: ea8d349d

packages/giselle-engine/src/core/generations/generate-text.ts
```diff
@@ -73,6 +77,23 @@ export async function generateText(args: {
...
 
+	const trace = langfuse.trace({
+		sessionId: runningGeneration.id,
+		name: "ai.streamText",
+		metadata: args.telemetry?.metadata,
+	});
+	const span = trace.span({
+		name: "ai.streamText",
+		startTime: new Date(runningGeneration.queuedAt),
+	});
```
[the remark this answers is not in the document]
rindrics: we don't need these comment because Langfuse SDK provides document: https://langfuse.com/docs/sdk/typescript/guide#trace

packages/language-model/src/costs/pricing.ts
```diff
@@ -0,0 +1,63 @@
...
+		totalCost: Cost;
+	} | null;
+}
+
+export class TokenBasedCostCalculator implements CostCalculator {
+	constructor(
+		private readonly pricing: Record<string, { prices: ModelPrice[] }>,
+	) {}
+
+	calculateCost(modelId: string, usage: TokenUsage) {
```
shige: @Rindrics 

I'd like you to add test code for `calculateCost`.

packages/language-model/src/costs/pricing.ts
```diff
@@ -0,0 +1,63 @@
+// Model cost in USD
+export type Cost = number;
+
+export function tokensToMegaTokens(tokens: number): number {
+	return tokens / 1_000_000;
```
shige, line 5: ### Findings and Recommendations on Floating-Point Calculations

The review of the pull request and relevant floating-point calculations reveals the following:

---

### Key Areas Involving Floating-Point Calculations:
1. **Cost Calculations in `packages/language-model/src/costs/pricing.ts`**:
   - **Code**: The `tokensToMegaTokens` function converts token counts to "mega tokens" (1,000,000 tokens).
   ```typescript
   export function tokensToMegaTokens(tokens: number): number {
       return tokens / 1_000_000;
   }
   ```
   - **Risk**: Division operations may introduce precision errors, especially for very large or very small token values. For example, `tokens = 999_999` would result in a fractional value when divided.
   - **Mitigation**: Consider rounding results to a fixed precision if exact values are required, e.g., `toFixed(6)`.

2. **Cost Calculation in `calculateTokenCost`**:
   - **Code**: This function computes costs using floating-point arithmetic:
   ```typescript
   const inputCost = tokensToMegaTokens(usage.input) * pricing.input.costPerMegaToken;
   const outputCost = tokensToMegaTokens(usage.output) * pricing.output.costPerMegaToken;
   ```
   - **Risk**: Multiplication of floating-point numbers can accumulate errors, especially when `tokensToMegaTokens` outputs very small numbers.
   - **Mitigation**:
     - Use libraries like `decimal.js` for precise arithmetic.
     - Ensure that all intermediate results are rounded or truncated appropriately.

3. **Usage Calculators (e.g., `TokenBasedCostCalculator`)**:
   - Similar risks and mitigations apply as in the points above. These calculators involve direct floating-point operations.

---

### General Observations:
- Floating-point operations are distributed across multiple files (e.g., `pricing.ts`, `calculator.ts`, `model-prices.ts`), primarily in cost and usage calculations.
- These calculations involve multiplication and division, which are prone to precision issues in JavaScript/TypeScript.

### Recommendations:
1. **Validation and Testing**:
   - Add unit tests for edge cases such as:
     - Very large token counts.
     - Very small token counts (e.g., `1` or `10`).
     - Rounding consistency in cost calculations.
   - Ensure test coverage for scenarios where operations might lose precision.

2. **Precision Control**:
   - Use a fixed precision for results where possible (e.g., `toFixed()` or `Number.toPrecision()`).
   - Consider using libraries like `big.js` or `decimal.js` for calculations requiring high precision.

3. **Error Handling**:
   - Validate and sanitize inputs to prevent unexpected behavior (e.g., negative or zero token counts).

---

### Further Action:
The pull request includes several modified and new files. To ensure a comprehensive review, I recommend focusing on:
- `packages/language-model/src/costs/pricing.ts`
- `packages/language-model/src/costs/calculator.ts`
- `packages/language-model/src/costs/model-prices.ts`

Let me know if you'd like detailed feedback on any specific section!

packages/language-model/src/costs/pricing.ts
```diff
@@ -0,0 +1,63 @@
...
+	constructor(
+		private readonly pricing: Record<string, { prices: ModelPrice[] }>,
+	) {}
+
+	calculateCost(modelId: string, usage: TokenUsage) {
+		const pricing = this.pricing[modelId]?.prices[0]?.price;
+		if (!pricing) return null;
+
+		const inputCost =
+			tokensToMegaTokens(usage.input) * pricing.input.costPerMegaToken;
```
shige: **Example of a floating-point problem**
```js
// usage.input = 300
// pricing.input.costPerMegaToken = 0.1

// tokensToMegaTokens(usage.input)
300 / 1_000_000
> 0.0003

// tokensToMegaTokens(usage.input) * pricing.input.costPerMegaToken
> 0.0003 * 0.1
0.000029999999999999997
```

shige: **Example of converting to an integer value and then calculating: (without using libraries such as big.js)**

```js
> ( 300 * (0.1 * 100) ) / (100 * 1_000_000)
0.00003

> ( 300 * (0.1 * 1_000_000) ) / (1_000_000 * 1_000_000)
0.00003
```

shige (changes requested):
There seems to be an issue with Floating-Point Calculations, please fix it!

packages/language-model/src/costs/pricing.ts
rindrics: fixed: https://github.com/giselles-ai/giselle/pull/834/commits/9f1a818889819b664cb2ccda2ebb15397518ef15

packages/language-model/src/costs/pricing.ts
rindrics: Thank you, I added tests to `calculateCost()` by commits after https://github.com/giselles-ai/giselle/pull/834/commits/13eb092d6af43862c8d1d3be7b448205604579e3

packages/language-model/src/costs/pricing.ts
rindrics: You're right.
We need another PR focusing on cost calculation logic.

However, I think we can merge this PR because:
- the calculation errors were avoided using integer-conversion method https://github.com/giselles-ai/giselle/pull/834/commits/9f1a818889819b664cb2ccda2ebb15397518ef15
- we don't bill users by model-usage at this point
  - we will do this after implementing robust & clear cost-calculation logic

@shige

packages/language-model/src/costs/calculator.ts
```diff
@@ -0,0 +1,59 @@
...
+	const outputCostPerMegaTokenInCents =
+		pricing.output.costPerMegaToken * dollarToCent * tokensPerMegaToken;
+
+	// Calculate costs in cents using integer arithmetic
+	const inputCostInCents = usage.promptTokens * inputCostPerMegaTokenInCents;
+	const outputCostInCents =
+		usage.completionTokens * outputCostPerMegaTokenInCents;
+	const totalCostInCents = inputCostInCents + outputCostInCents;
+
+	// Convert back to dollars by dividing by 1,000,000 (mega tokens) and 100 (cents)
```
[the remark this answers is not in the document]
rindrics: I'll remain the comment as it is because I think the suggested version is too much

rindrics:
📝 we should group trace by workspace ID not generation ID

packages/language-model/src/costs/pricing.ts
shige: @Rindrics I confirmed that calculateTokenCost has been fixed, but calculateCost does not seem to have been fixed yet.
https://github.com/giselles-ai/giselle/pull/834/files#diff-1b0de3799947b0b3eabd115b38249c31c721490baa840d97eb68051cbb6c5763R47

packages/language-model/src/costs/calculator.ts
```diff
@@ -0,0 +1,59 @@
...
+export class DefaultCostCalculator implements CostCalculator {
+	constructor(private provider: string) {}
+
+	async calculate(model: string, usage: ModelUsage): Promise<CostResult> {
+		console.log(`No cost calculator found for ${this.provider}`);
+		return { input: 0, output: 0, total: 0 };
+	}
+}
+
+export function calculateTokenCost(
```
shige: IMO:

I think the codes and comments are difficult to understand, so I would like you to refactor it, even if it's a separate issue or PR.

shige:
@toyamarinyon @satococoa 
Since this will be the base implementation for Giselle's fee calculation, I would like to ask all developers to review this PR!

packages/language-model/src/costs/pricing.ts
rindrics: sorry, I deleted it by ae0dce90 because it was not referenced @shige

packages/giselle-engine/src/core/generations/generate-text.ts
```diff
@@ -78,6 +85,23 @@ export async function generateText(args: {
...
+	const generation = span.generation({
+		name: "ai.streamText.doStream",
+		model: operationNode.content.llm.id,
+		modelParameters: operationNode.content.llm.configurations,
+		startTime: new Date(runningGeneration.createdAt),
+		completionStartTime: new Date(runningGeneration.startedAt),
+	});
+
```
toyamarinyon: Why do you need to directly operate the Langfuse instance? I would like to know the reason why it cannot be done within the framework of AI SDK Telemetry.
https://ai-sdk.dev/docs/ai-sdk-core/telemetry

Also, if it cannot be done within the framework of AI SDK Telemetry, is it not possible to create a Custom Tracer to address this?
https://ai-sdk.dev/docs/ai-sdk-core/telemetry#custom-tracer

packages/giselle-engine/src/core/generations/generate-text.ts
```diff
@@ -78,6 +85,23 @@ export async function generateText(args: {
 		startedAt: Date.now(),
 	} satisfies RunningGeneration;
 
+	const trace = langfuse.trace({
+		userId: String(args.telemetry?.metadata?.userId),
+		name: "ai.streamText",
```
satococoa: Is this a recommended or default naming by Langfuse?

packages/giselle-engine/src/core/generations/generate-text.ts
```diff
@@ -78,6 +85,23 @@ export async function generateText(args: {
...
+		metadata: args.telemetry?.metadata,
+	});
+	const span = trace.span({
+		name: "ai.streamText",
+		startTime: new Date(runningGeneration.queuedAt),
```
satococoa: I wonder which is better, runningGeneration.queuedAt or runningGeneration.startedAt?
I'd like to get your thoughts on using runningGeneration.queuedAt versus runningGeneration.startedAt.

packages/giselle-engine/src/core/generations/generate-text.ts
```diff
@@ -396,6 +434,31 @@ export async function generateText(args: {
...
+			updateTelemetry(span, { messages }, event.text);
+			updateTelemetry(generation, { messages }, event.text);
+			const tokenUsage = event.usage;
+
+			let costInfo = null;
+			const provider = operationNode.content.llm.provider;
+			const modelId = operationNode.content.llm.id;
+
+			if (tokenUsage) {
+				const { calculateCost } = await import("@giselle-sdk/language-model");
```
satococoa: Can we use static(normal) imports instead of dynamic imports in this module?

packages/giselle-engine/src/core/generations/telemetry.ts
```diff
@@ -1,16 +1,35 @@
...
+
+type TelemetryTag =
+	| BaseFunctionalityTag
+	| ProviderNameTag
+	| ProviderOptionTag
+	| ModelNameTag;
+
+let langfuseInstance: Langfuse | null = null;
+
+export function getLangfuseInstance(): Langfuse {
```
satococoa: Do we need to implement this as a Singleton? If new Langfuse() is a heavy operation, this approach makes sense. But if it's not resource-intensive, creating new instances would prevent unintended state sharing.

packages/giselle-engine/src/core/generations/generate-text.ts
```diff
@@ -78,6 +85,23 @@ export async function generateText(args: {
 		startedAt: Date.now(),
 	} satisfies RunningGeneration;
 
+	const trace = langfuse.trace({
+		userId: String(args.telemetry?.metadata?.userId),
```
satococoa: Since teams are our billing entities, I'm wondering why we need userId in the telemetry data. Wouldn't teamId alone be sufficient for billing purposes?

packages/giselle-engine/src/core/generations/generate-text.ts
rindrics: > Why do you need to directly operate the Langfuse instance?

we need Langfuse instance because [`experimental_telemetry` of `streamText()` does not have attribute`cost`](https://ai-sdk.dev/docs/reference/ai-sdk-core/stream-text#experimental_telemetry.telemetry-settings).

> is it not possible to create a Custom Tracer to address this?
> https://ai-sdk.dev/docs/ai-sdk-core/telemetry#custom-tracer

yes, I think we can use our own custom tracer (not `tracerProvider.getTracer('ai')` shown in the reference) in the future.
Let me try this in another PR.

packages/language-model/src/costs/calculator.ts
```diff
@@ -0,0 +1,59 @@
...
+		pricing.output.costPerMegaToken * dollarToCent * tokensPerMegaToken;
+
+	// Calculate costs in cents using integer arithmetic
+	const inputCostInCents = usage.promptTokens * inputCostPerMegaTokenInCents;
+	const outputCostInCents =
+		usage.completionTokens * outputCostPerMegaTokenInCents;
+	const totalCostInCents = inputCostInCents + outputCostInCents;
+
+	// Convert back to dollars by dividing by 1,000,000 (mega tokens) and 100 (cents)
+	const divisor = tokensPerMegaToken * dollarToCent * tokensPerMegaToken;
```
satococoa, line 60: The `Calculator` should ideally return precise values.

Returning a floating-point result here may introduce precision errors, which is an anti-pattern for handling money in JavaScript.

I recommend performing all calculations in integers and only converting to floating-point for display.

packages/giselle-engine/src/core/generations/generate-text.ts
toyamarinyon: Generation is the core of the core function in Giselle. Although you might think that telemetry could also be implemented in the same detailed way, I believe that telemetry should have a thin implementation that doesn't make its presence felt. (Alternatively, there could be a design that wraps it as a template function like in the AI SDK.) Looking forward to the implementation.

toyamarinyon (approved):

packages/language-model/src/costs/model-prices.ts
```diff
@@ -0,0 +1,178 @@
+import type { ModelPrice } from "./pricing";
+
+export type ModelPriceTable = Record<string, { prices: ModelPrice[] }>;
+
+const validPriceCache = new Map<string, ModelPrice>();
```
satococoa: Is implementing a cache necessary here?
Directly retrieving the current valid price seems adequate and less error-prone.

packages/language-model/src/costs/model-prices.ts
```diff
@@ -0,0 +1,178 @@
...
+
+let lastCacheResetDate = new Date();
+
+// Check and reset cache if it's a new UTC day
+function checkAndResetCacheIfNeeded() {
```
satococoa: We would have to consider time precision (hour, minutes, seconds) if we implement a cache mechanism.

satococoa (changes requested):
I've added some comments!

packages/giselle-engine/src/core/generations/generate-text.ts
rindrics: > I believe that telemetry should have a thin implementation that doesn't make its presence felt

I agree, it's always true!
I'll refactor them into the form they should be.

packages/giselle-engine/src/core/generations/generate-text.ts
rindrics: no, we should reconsider their (trace, span, generation) naming convention in the future.

packages/giselle-engine/src/core/generations/generate-text.ts
rindrics: Thanks, I set `runningGeneration.startedAt` to the start time of `generation` (not `span`, focused here).

I think my configuration is reasonable, but there is a room to discuss this if we have to improve speed performance of generation node.
Please let me leave this configuration as it is in this PR.

packages/giselle-engine/src/core/generations/generate-text.ts
rindrics: Thanks, I simplified the code using static import by 3f1994c4

packages/giselle-engine/src/core/generations/telemetry.ts
rindrics: I thought sharing Langfuse instance itself as singleton is OK because every generation creates new trace instance by

```node
const trace = langfuse.trace();
```

But after reading your comment I feel I should reconsider about this

because I couldn't find articles that suggests to create Langfuse instance as a Singleton 🤔

packages/giselle-engine/src/core/generations/generate-text.ts
toyamarinyon: I wonder if it wasn't using "await import" because it was necessary? Why was it initially implemented with "await import"?

I'm curious about the background since I don't think it would have ended up this way with a normal implementation.

packages/giselle-engine/src/core/generations/telemetry.ts
rindrics: I will revert edfa68a1e46ad24a0f08258cfa2816bee2b76b73

packages/giselle-engine/src/core/generations/generate-text.ts
rindrics: > Wouldn't teamId alone be sufficient for billing purposes

Yes, it is sufficient for the billing purposes.

This is for an analytical purpose: Langfuse provides functionality to track usage of single user
So let me leave the code as it is 👍 

The user in the screenshot below is me on the development environment:
image

packages/language-model/src/costs/calculator.ts
rindrics: Thanks,

> I recommend performing all calculations in integers and only converting to floating-point for display.

This is great if we could this "always", but in this situation we can't.
Because the cost of single generation will be in float if we adopt cents for our calculation unit:

For example,
- cost of "mega" token is 200 cents (2.0 dollar) in gpt-4.1
  - so if the tokens consumed by a single generation is less than 5000, the calculated cost in cents will be smaller than 1 cents
  - > if the tokens consumed by a single generation is less than 5000
    - and this happens so often

Please let me leave the code here as it is because we already avoided the calculation errors were avoided using integer-conversion method https://github.com/giselles-ai/giselle/commit/9f1a818889819b664cb2ccda2ebb15397518ef15

packages/language-model/src/costs/model-prices.ts
rindrics: Yes, I'll consider the speed-related performance solution "after" speed-related problem occurred.
We should focus on accuracy first.

rindrics: I'll revert ea8d349da81619663831e5b4c6ad04d7aca4269e

packages/giselle-engine/src/core/generations/generate-text.ts
satococoa: got it! 👍

packages/giselle-engine/src/core/generations/generate-text.ts
rindrics: Sorry, it was not by necessity.

The first version simply loaded the function only when `tokenUsage` was non-null,
but because the function is not heavy, moreover, dynamic imports are basically not used in this repository, I updated this line to use normal import. @toyamarinyon

packages/language-model/src/costs/model-prices.ts
rindrics: reverted by a97de5ad

packages/language-model/src/costs/model-prices.ts
rindrics: Thank you.
This comment became outdated because I quit caching by a97de5ad

packages/giselle-engine/src/core/generations/telemetry.ts
rindrics: reverted by c62d7e5d

rindrics:
Thank you, I replied to your each comment.
Please let me know if you still have any questions @satococoa

packages/language-model/src/costs/calculator.ts
satococoa: > the cost of single generation will be in float

I see.

Since we offer monthly billing, the amount of model usage fees is a sum of these 'costs of single generation'.
For display purposes of a single generation, this float value is sufficient and rounding is fine.
However, we cannot use these rounded values to calculate the sum of monthly model usage accurately.

Is this logic and return value only used for display purposes, or do you plan to use this value to calculate monthly fees as well?

----

For example, here are my ideas about how to calculate "monthly fees":

## Plan A: Record calculated price in sufficient precision integers

Instead of using floating-point numbers, we can decide on an appropriate level of precision (e.g., millicents, microcents) and store all prices as integers.

For example:
- $1.25 would be stored as 1,250,000 microcents
- $0.00123 would be stored as 1,230 microcents

This approach:
- Avoids floating-point errors completely
- Allows exact summing of values
- Requires only one division operation when displaying the final amount
- Is compatible with most database systems

## Plan B: Record token value

In this plan, rounding is only occur only once when calculating monthly fees.

satococoa: Since calculation monthly fees is out of this pull request's scope, I want to know here is only this: 🙏 

> Is this logic and return value only used for display purposes, or do you plan to use this value to calculate monthly fees as well?

rindrics: Ah, I finally understand what you meant 🙇 

Item | Option A (Store cost in millicents) | Option B (Store token count only)
-- | -- | --
Implementation Simplicity | Calculate and store cost per request | Store only token count, calculate later
Flexibility (Rate Changes) | Low (hard to adjust past records) | High (can recalculate past costs easily)
Accuracy (Rounding Errors) | Very high (integer-based) | Requires care to avoid rounding errors
Billing Aggregation Logic | Simple (just sum stored amounts) | More complex (match token usage with rates)
Storage Efficiency | Lower (store cost every time) | Higher (store only token usage)

I suggest `Option B`.
The drawback of the option, "Requires care to avoid rounding errors", won't be a problem if we clarify the rounding logic (on the monthly batch) by document or FAQ

What do you think? @satococoa cc: @shige

rindrics: > Is this logic and return value only used for display purposes, or do you plan to use this value to calculate monthly fees as well?

Firstly I wrote with the intention of fee calculations, but now I realized that at least with current implementation, the calculated value should be used for "display only"! @satococoa

satococoa: > I suggest Option B.

Good 👍 
And I understand this method is intended to be used for "display only".

packages/giselle-engine/src/core/generations/telemetry.ts
```diff
@@ -1,48 +1,160 @@
...
+			unit,
+		},
+		startTime: new Date(runningGeneration.createdAt),
+		completionStartTime: new Date(runningGeneration.startedAt),
+		metadata: settings?.metadata,
+		output,
+		endTime: new Date(completedGeneration.completedAt),
+	});
+
+	return { langfuse, trace, span, generation };
```
rindrics, line 163: need to return not only Langfuse object, but also other child tracer objects to get telemetry data correctly sent to Langfuse

rindrics:
From your last review, I updated the code in the points below:
- explain that cost calculation function is for "display-only" f3509425
- minimize exposure of telemetry codes in `generate-text.ts` b26e169d

would you review the code again? @satococoa

packages/giselle-engine/src/core/generations/generate-text.ts
```diff
@@ -469,6 +484,37 @@ export async function generateText(args: {
 				generation: completedGeneration,
 				onConsumeAgentTime: args.context.onConsumeAgentTime,
 			});
+
+			console.log("before---------------");
```
satococoa:
```suggestion
```

packages/giselle-engine/src/core/generations/generate-text.ts
```diff
@@ -469,6 +484,37 @@ export async function generateText(args: {
...
+					outputCost: costInfo?.outputCost ?? 0,
+					totalCost: costInfo?.totalCost ?? 0,
+				},
+				completedGeneration,
+				spanName: "ai.streamText",
+				generationName: "ai.streamText.doStream",
+				unit: "TOKENS",
+				settings: args.telemetry,
+			});
+			console.log("after---------------");
```
satococoa:
```suggestion
```

packages/giselle-engine/src/core/generations/telemetry.ts
```diff
@@ -59,3 +76,93 @@ export function generateTelemetryTags(args: {
...
+	settings?: TelemetrySettings;
+}): {
+	langfuse: Langfuse;
+	trace: ReturnType<Langfuse["trace"]>;
+	span: ReturnType<ReturnType<Langfuse["trace"]>["span"]>;
+	generation: ReturnType<
+		ReturnType<ReturnType<Langfuse["trace"]>["span"]>["generation"]
+	>;
+} {
+	console.log("inside======");
```
satococoa: Are leaving these `console.log()`s OK for us? or should remove? cc @shige

packages/giselle-engine/src/core/generations/telemetry.ts
satococoa: It appears that the returned objects in `generate-text.ts` are not being utilized. 
Should we still return these objects, or would it be better to refactor this code?

packages/language-model/src/costs/index.ts
```diff
@@ -0,0 +1,40 @@
...
+	switch (provider) {
+		case "openai":
+			return new OpenAICostCalculator();
+		default:
+			console.log(`Unimplemented provider: ${provider}`);
+			return new DefaultCostCalculator(provider);
+	}
+}
+
+export async function calculateCost(
```
satococoa: I think this is a bit misleading.

Since the calculation method used in this logic is named `calculateTokenCostForDisplay`, hiding this intention could be problematic.

If it's acceptable that the price value sent to Langfuse is a display price, then the metric name and calculation method should be named something like `displayPrice`.

satococoa (changes requested):
Thank you!
I've added some additional comments 💬

packages/giselle-engine/src/core/generations/telemetry.ts
rindrics: sorry, I removed it by e198bc2a 🙇

packages/giselle-engine/src/core/generations/generate-text.ts
rindrics: fixed: e198bc2a

packages/giselle-engine/src/core/generations/generate-text.ts
rindrics: fixed: e198bc2a

packages/giselle-engine/src/core/generations/telemetry.ts
rindrics: > Should we still return these objects

yes, need to return to get telemetry sent to Langfuse!

> or would it be better to refactor this code?

yes, I renamed these objects with adding prefix `_`: a18b17e0

packages/language-model/src/costs/index.ts
rindrics: Thank you.
I renamed objects that may mislead the their intention: d602f5db

packages/giselle-engine/src/core/generations/generate-text.ts
```diff
@@ -469,6 +484,41 @@ export async function generateText(args: {
...
 				onConsumeAgentTime: args.context.onConsumeAgentTime,
 			});
+
+			// necessary to send telemetry but not explicitly used
+			const {
+				langfuse: _langfuse,
+				trace: _trace,
+				span: _span,
+				generation: _generation,
+			} = createLangfuseTracer({
```
satococoa: Nit: If we need to keep these objects for telemetry, maybe we could write the code like this instead:

```suggestion
			const _langfuseTracer = createLangfuseTracer({
```

satococoa (approved):
Thank you!

I think the naming and structure of price calculation functions have room for refinement, but it would be better to restructure them when we implement monthly pricing.

packages/giselle-engine/src/core/generations/generate-text.ts
rindrics: Thank you, I confirmed that the telemetries arrived after applying the change suggestion above! ✅

rindrics:
We decided to implement cost calculation for "billing" in https://github.com/giselles-ai/giselle/issues/865.
Now we can merge this PR because it calculates cost for displaying preliminary value @shige

rindrics:
Thank you for your reviews!

PR #847, opened 2025-05-14
remove toyamarinyon from codeowners

toyamarinyon:
I removed myself as a Code Owner because I felt my comments as a Code Owner were not contributing to the product. I would appreciate if you could assign me as a reviewer when it seems necessary.


shige (approved):
>  I would appreciate if you could assign me as a reviewer when it seems necessary.

I got it, thank you!

Issue #815, opened 2025-05-08
Tag Langfuse trace with provider tool configs


Rindrics:
## Task description

Tag Langfuse trace with provider tool configs to record app usage

## Background

Giselle provides text-generation nodes with provider configs such as:
- web-search (OpenAI)
- reasoning (Anthropic)
- search grounding (Google)

We should tag Langfuse trace with these information to get observability of our app

rindrics:
- done by #816

PR #848, opened 2025-05-14
Add tags to Langfuse trace to make model usage data analyzable

Rindrics:
## Summary

Add tags below to Langfuse trace:
- user ID
- provider name
- model name

## Related Issue

We already tagged traces with provider tool config by #815, but we need more analytical clues to make usage data analyzable

## Testing

New tags are observed:
image


shige (approved):
Thank you! 🆙 

---

📝 For My Refs
- [Users \- Langfuse](https://langfuse.com/docs/tracing-features/users)

rindrics:
thanks!

PR #859, opened 2025-05-19
Add copyItem method to blob-storage

toyamarinyon:
## Summary
- extend `BlobStorageDriver` interface with optional `copyItem`
- implement `copyItem` for filesystem and Supabase drivers

## Testing
- `npx biome check packages/blob-storage --write`
- `pnpm test` *(fails: corepack cannot download pnpm during studio.giselles.ai tests)*


toyamarinyon:
I'll try again later

PR #868, opened 2025-05-19
Revert "Revert "Improve workflow node connections and generation performance""

toyamarinyon:
In PR #861, we added an accessor to Input Document. While we implemented data migration, there were special cases where the migration was missed, so we addressed this in commit 1c95a6a.

The root cause is that parts of the API types are set to 'Any', but addressing this properly would require API type generation which would impact development speed. We're considering better approaches to handle this.

There's also another issue where we want to perform data migration at deployment time rather than ad-hoc, but even this approach wouldn't have prevented the current problem.


shige (approved):
Thank you! 👍

toyamarinyon:
Potential issue is https://github.com/giselles-ai/giselle/issues/869

PR #871, opened 2025-05-19
Include cost (for display not billing) to Langfuse

Rindrics:
## Summary

Fix construction of telemetry data to send cost info to Langfuse correctly

## Related Issue

- #834 calculated the cost info but failed to include them in telemetry

## Testing

Cost info of OpenAI model arrived ✅
image

Cost info of Anthropic model is blank as expected ✅ 
- Because their unit price is not implemented yet (will be added by #872)

image

But in auto-instrumented telemetry, cost is ["inferred" by usage](https://langfuse.com/docs/model-usage-and-cost#infer)

### DIFF
--- a/packages/giselle-engine/src/core/generations/telemetry.ts
+++ b/packages/giselle-engine/src/core/generations/telemetry.ts
@@ -148,11 +148,7 @@ export function createLangfuseTracer({
 		modelParameters:
 			runningGeneration.context.operationNode.content.llm.configurations,
 		input: messages,
-		usage: {
-			input: usage.input,
-			output: usage.output,
-			unit,
-		},
+		usage,
 		startTime: new Date(runningGeneration.createdAt),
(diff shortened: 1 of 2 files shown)


packages/giselle-engine/src/core/generations/telemetry.ts
```diff
@@ -148,11 +147,7 @@ export function createLangfuseTracer({
 		modelParameters:
 			runningGeneration.context.operationNode.content.llm.configurations,
 		input: messages,
-		usage: {
-			input: usage.input,
-			output: usage.output,
-			unit,
-		},
+		usage,
```
rindrics, line 150: Let me fix this because the original code discarded cost info

shige:
I'd like to have one of the @giselles-ai/giselle-developers review it!

shige (approved):
Thank you! 👍

rindrics:
Thank you for your review!

PR #873, opened 2025-05-19
Refactor: Use type-only imports and Proxy-based client for router paths

toyamarinyon:
This PR changes the imports of formDataRouterPaths and jsonRouterPaths to be type-only imports to avoid bringing in server implementations. A Proxy-based approach is used to dynamically create API method handlers when they're accessed.

## Changes

- Changed the imports to be type-only imports
- Implemented a Proxy-based approach that dynamically creates methods when they're accessed
- The Proxy's get trap intercepts property access and returns a function that makes the API request
- FormData endpoints are determined by checking if the property name is 'uploadFile'
- This approach completely avoids the need to import and iterate over the path arrays at runtime

### DIFF
--- a/packages/giselle-engine/src/react/use-giselle-engine.ts
+++ b/packages/giselle-engine/src/react/use-giselle-engine.ts
@@ -1,14 +1,12 @@
 import { useCallback, useMemo } from "react";
 import type { z } from "zod";
-import {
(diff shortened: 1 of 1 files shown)


shige (approved):
Thank you! ♻️

PR #890, opened 2025-05-20
Set environment name and deployment ID to Langfuse trace

Rindrics:
## Summary

Set information below to Langfuse trace make their development easier:
- name of environment 
- Vercel deployment ID

## Testing

"Environment" is set to the trace arrived to Langfuse ✅ 
image

deployment ID is set as metadata ✅ 
image

### DIFF
--- a/packages/giselle-engine/src/core/generations/telemetry.ts
+++ b/packages/giselle-engine/src/core/generations/telemetry.ts
@@ -120,7 +120,15 @@ export function createLangfuseTracer({
 	const trace = langfuse.trace({
 		userId: String(settings?.metadata?.userId),
 		name: spanName,
-		metadata: settings?.metadata,
+		metadata: {
+			...settings?.metadata,
+			...(process.env.VERCEL_DEPLOYMENT_ID && {
+				deploymentId: process.env.VERCEL_DEPLOYMENT_ID.replace(
+					"dpl_",
+					"",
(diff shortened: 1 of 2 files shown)


packages/giselle-engine/src/core/generations/telemetry.ts
```diff
@@ -120,7 +120,15 @@ export function createLangfuseTracer({
 	const trace = langfuse.trace({
 		userId: String(settings?.metadata?.userId),
 		name: spanName,
-		metadata: settings?.metadata,
+		metadata: {
+			...settings?.metadata,
+			...(process.env.VERCEL_DEPLOYMENT_ID && {
+				deploymentId: process.env.VERCEL_DEPLOYMENT_ID.replace(
+					"dpl_",
```
rindrics, line 127: - remove "dpl_"
- use first 9 letters

to format raw deployment ID to the one on Vercel .com
image

image

shige: @Rindrics 

If the format of `VERCEL_DEPLOYMENT_ID` changes, we will have to respond, so I thought it would be okay not to replace it.
Since the prefixes used in Giselle DB, such as `usr_`, are still used, I thought it would be fine to leave `dpl_` here as well.

rindrics: sorry, I unintentionally closed this PR (by renaming "remote" `langfuse-tag-env` branch), but I'll correct this in another PR @shige 
Thank you!

PR #943, opened 2025-05-23
Refactor GitHub event reactions

toyamarinyon:
## Summary
- refactor GitHub event handlers to return a node ID for reactions instead of calling `addReaction`
- invoke `addReaction` from `processEvent`
- update tests for new behaviour

## Testing
- `pnpm -F @giselle-sdk/giselle-engine test`


shige (approved):
Thank you! ♻️

PR #952, opened 2025-05-26
Add pull request diff retrieval

toyamarinyon:
## Summary
- implement `getPullRequestDiff` for Github Tool
- export new function from package index

## Testing
- `npx turbo format --cache=local:rw`
- `npx turbo check-types --cache=local:rw`
- `npx turbo test --cache=local:rw`


shige (approved):
Thank you! 🧰

PR #954, opened 2025-05-26
Log trigger info when processEvent fails

toyamarinyon:
## Summary
- log trigger workspaceId and nodeId when `processEvent` throws in webhook handler

## Testing
- `turbo test --cache=local:rw`

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved error handling to prevent interruptions when processing events. Errors are now logged, allowing other processes to continue without disruption.


shige (approved):
Thank you! 🔍

PR #960, opened 2025-05-27
Add migration files to save model usage

Rindrics:


apps/studio.giselles.ai/drizzle/schema.ts
```diff
@@ -333,3 +333,32 @@ export const githubRepositoryEmbeddings = pgTable(
...
+export type UsageUnit = "tokens";
+
+export const modelUsageItems = pgTable("model_usage_items", {
+	dbId: serial("db_id").primaryKey(),
+	modelUsageDbId: integer("model_usage_db_id")
+		.notNull()
+		.references(() => modelUsage.dbId, { onDelete: "cascade" }),
+	usageMetric: text("usage_metric").$type<UsageMetric>().notNull(),
+	amount: integer("amount").notNull(),
+	unit: text("unit").$type<UsageUnit>().notNull(),
```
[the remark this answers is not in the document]
rindrics: I'll leave the code as it is since it can benefit from TypeScript type checking

apps/studio.giselles.ai/drizzle/schema.ts
```diff
@@ -333,3 +333,46 @@ export const githubRepositoryEmbeddings = pgTable(
...
+	{
+		dbId: serial("db_id").primaryKey(),
+		modelUsageDbId: integer("model_usage_db_id")
+			.notNull()
+			.references(() => modelUsage.dbId, { onDelete: "cascade" }),
+		usageMetric: text("usage_metric").$type<UsageMetric>().notNull(),
+		amount: integer("amount").notNull(),
+		unit: text("unit").$type<UsageUnit>().notNull(),
+		endedAt: timestamp("ended_at").notNull(),
+		createdAt: timestamp("created_at").defaultNow().notNull(),
```
satococoa, line 382: I have questions: 


Q1. Do we need separate timestamp columns in the `model_usage_items` table?

We have `model_usage.created_at`, is it sufficient?

Q2. Do `endedAt` and `createdAt` in the `model_usage_items` table always have the same timestamp?

satococoa (commented):
Thank you!
I've added a comment. Could you take a look?

apps/studio.giselles.ai/drizzle/schema.ts
```diff
@@ -333,3 +333,46 @@ export const githubRepositoryEmbeddings = pgTable(
...
 	}),
 );
+
+export const modelUsage = pgTable(
+	"model_usage",
+	{
+		dbId: serial("db_id").primaryKey(),
+		teamDbId: integer("team_db_id")
+			.notNull()
+			.references(() => teams.dbId, { onDelete: "cascade" }),
```
shige, line 349: Please add `userDbId` so we can identify the executor. I think it's okay to allow a null value if it's executed via a Trigger Node.

apps/studio.giselles.ai/drizzle/schema.ts
```diff
@@ -333,3 +333,46 @@ export const githubRepositoryEmbeddings = pgTable(
 		),
 	}),
 );
+
+export const modelUsage = pgTable(
```
shige:
```suggestion
export const modelUsages = pgTable(
```

apps/studio.giselles.ai/drizzle/schema.ts
```diff
@@ -333,3 +333,46 @@ export const githubRepositoryEmbeddings = pgTable(
...
+	"model_usage",
```
shige:
```suggestion
	"model_usages",
```

apps/studio.giselles.ai/drizzle/schema.ts
rindrics: > Q1. Do we need separate timestamp columns in the model_usage_items table?
> We have model_usage.created_at, is it sufficient?

Thank you, `model_usage.created_at` seems to be sufficient.

> Q2. Do endedAt and createdAt in the model_usage_items table always have the same timestamp?

no, endedAt will be injected from app; something like `endedAt: completedGeneration.completedAt`.

On the other hand, `createdAt` is the timestamp of the DB record creation.

rindrics:
> We maintain the pricing table (ModelPriceTable) in code, but if we don’t store a price snapshot in the database, we risk being unable to recalculate historical usage fees.
> Please make model_usage or model_usage_items capable of referencing the ModelPriceTable in code.

Thank you.
I'd like to keep the current implementation as it is because I think we can already reference the pricing information through the following methods:
1. The combination of `model` and `provider` columns in `model_usage` table directly maps to the keys in our `ModelPriceTable`, allowing us to reference the pricing information.
1. We can determine the applicable price for any usage record by comparing the `endedAt` timestamp in `model_usage_items` with the `validFrom` dates in our `ModelPriceTable`

@shige

apps/studio.giselles.ai/drizzle/schema.ts
rindrics: fixed: 6331e59b

shige:
@Rindrics cc/ @satococoa 

> 1\. The combination of `model` and `provider` columns in `model_usage` table directly maps to the keys in our `ModelPriceTable`, allowing us to reference the pricing information.

Do these keys necessarily match those of `ModelPriceTable`?
If so, there's no problem, but they could not be read from the existing Table definition.

---

> 2\. We can determine the applicable price for any usage record by comparing the `endedAt` timestamp in `model_usage_items` with the `validFrom` dates in our `ModelPriceTable`

By comparing the endedAt timestamp in model_usage_items with the validFrom date in ModelPriceTable, we can determine the price that applies to the usage record.

Currently, ModelPriceTable is managed in source code, so the date and time defined in the validFrom date may not have been released to production.

For this reason, I felt that a simple comparison of time information would not be sufficient.

---

FYI: It may be a good idea to imagine the work of customer support and support engineers when designing what values ​​should be stored in the database.

For example, if someone asks, "I don't know the breakdown of the $123 usage fee on April 2, 2025, please tell me,"
you would need to reverse-look up the LLM API call log, database records, pricing calculation logic, etc.

With the current implementation, the data stored in the database does not meet the minimum requirements, and it is easy to imagine that supporting customers would become very time-consuming.

satococoa:
I agree.

Having to search for which price is applied to a specific model_usage will be time-consuming and fragile.
Let's have a column that saves which price is applied.

----

Add `ModelPrice` to price_id like this:

https://github.com/giselles-ai/giselle/blob/4d4e0a2c305b575643cca06a37d49c207de7ab49/packages/language-model/src/costs/pricing.ts#L21-L24

```ts
export type ModelPrice = {
        id: ModelPriceId; // <- add this
	validFrom: string;
	price: TokenBasedPricing;
};
```

And save this value into ModelUsages.priceId (string with `ModelPriceId` type)

This way, we can avoid expensive lookups and ensure data consistency.
What do you think? > @Rindrics @shige

shige:
That's a good idea! 
I think it will resolve my concerns because it will be unique. 👍

rindrics:
> > 1\. The combination of `model` and `provider` columns in `model_usage` table directly maps to the keys in our `ModelPriceTable`, allowing us to reference the pricing information.
> 
> Do these keys necessarily match those of `ModelPriceTable`?
> If so, there's no problem, but they could not be read from the existing Table definition.

Thanks, the consistency between these values are not ensured  at that point.
I fixed this problem by referring model names using constant in https://github.com/giselles-ai/giselle/pull/960/commits/d030a257b592f19e723fdc4b07644608a00642bc 👍 

> With the current implementation, the data stored in the database does not meet the minimum requirements, and it is easy to imagine that supporting customers would become very time-consuming.

That's right. It was an important suggestion...
Also applying advice from @satococoa, I added a column `priceId` to the table  `model_usages` （https://github.com/giselles-ai/giselle/pull/960/commits/105e1bbec3743e3f0a710eac7e52aaf21cd3fff8 ）

@shige

packages/language-model/src/models.ts
```diff
@@ -0,0 +1,35 @@
...
+	O3: "o3",
+	O3_MINI: "o3-mini",
+	O4_MINI: "o4-mini",
+	GPT4_1: "gpt-4.1",
+	GPT4_1_MINI: "gpt-4.1-mini",
+	GPT4_1_NANO: "gpt-4.1-nano",
+} as const;
+
+export const ANTHROPIC_MODELS = {
+	CLAUDE4_OPUS: "claude-4-opus-20250514",
```
shige, line 15:
```suggestion
	"claude-4-opus-20250514": "claude-4-opus-20250514",
```

shige: IMO:

Even if we were to make it a constant, I thought it would be easier to understand if we made the key-value the same value.

apps/studio.giselles.ai/drizzle/schema.ts
```diff
@@ -5,6 +5,12 @@ import type {
 } from "@/services/external/github/types";
 import type { TeamId } from "@/services/teams/types";
 import type { WorkspaceId } from "@giselle-sdk/data-type";
+import type {
+	AnthropicModel,
+	GoogleModel,
+	ModelPriceId,
+	OpenAIModel,
+} from "@giselle-sdk/language-model";
```
shige, line 13: Are `AnthropicModel`, `GoogleModel` and `OpenAIModel` unused?

packages/language-model/src/models.ts
shige: I'd like to hear your opinion. @satococoa @toyamarinyon

shige:
Since the scope of the modifications extends to `packages/language-model/src/models.ts`, I would like to ask @toyamarinyon san to review it as well.

packages/language-model/src/models.ts
satococoa: Using model ID as the key is good.
But we should prepare for this case:

- e.g., `claude-4-opus-20250514` changes its price without adding another date suffix

Therefore, I would like to recommend the following approach:
- Adding a sequential number to model ID or provider.
  - `"claude-4-opus-20250514-1"` or `"anthropic-1"`

toyamarinyon: Before discussing the merits of this expression/approach, I want to know why this change is necessary for achieving this Pull request's objective of 'Add DB schema to save model usage'.

toyamarinyon (commented):
At what granularity do you want to record and measure usage? I read the schema, code, and related issues, but I couldn't understand it well.

rindrics:
> At what granularity do you want to record and measure usage?

We have to record tokens (not costs) consumed by each generation, such as the values included in the [response of AI SDK `streamText()`](https://ai-sdk.dev/docs/reference/ai-sdk-core/stream-text#on-step-finish.on-step-finish-result.usage). @toyamarinyon

toyamarinyon:
> We have to record tokens (not costs) consumed by each generation, such as the values included in the [response of AI SDK `streamText()`](https://ai-sdk.dev/docs/reference/ai-sdk-core/stream-text#on-step-finish.on-step-finish-result.usage).

@Rindrics  For this use case, I think it would be natural to add a usage object to Generation, but why don't you do that?

rindrics:
Records of model usage should be stored as a separate table because we're going to use them in later billing process @toyamarinyon

toyamarinyon:
@Rindrics I have two main concerns about the current structure:

1. **Too complex for the core requirement** - The current two-table approach seems over-engineered for saving model usage
2. **Cannot track individual generations** - The current structure doesn't allow us to identify which specific AI generation incurred which costs

### Issues with Current Structure

- **Generation tracking**: We can't trace costs back to individual AI generations, making debugging and optimization difficult
- **Unnecessary complexity**: Splitting into `model_usage` + `model_usage_items` creates complexity without clear benefits

### Proposed Structure

```mermaid
erDiagram
    model_prices {
        bigint db_id PK
        text price_id UK "Internal price ID (e.g., gpt-4o-2025-01-01)"
        text model_id "Model identifier (e.g., gpt-4o)"
        text provider "Provider name (e.g., openai)"
        decimal input_cost_per_mega_token "Cost per 1M input tokens"
        decimal output_cost_per_mega_token "Cost per 1M output tokens"
        timestamp valid_from "Price effective from"
        timestamp valid_until "Price effective until (NULL = indefinite)"
        boolean is_active "Active flag"
        timestamp created_at
        timestamp updated_at
    }

    generation_usage {
        bigint db_id PK
        integer team_db_id FK
        integer agent_db_id FK
        integer user_db_id FK "NULL for trigger executions"
        text generation_id "Individual generation identifier"
        text model_id "Model used (e.g., gpt-4o)"
        text provider "Provider (e.g., openai)"
        text price_id FK "Price used for calculation"
        integer input_tokens "Input token count"
        integer output_tokens "Output token count"
        decimal input_unit_cost "Input cost per token at time of use"
        decimal output_unit_cost "Output cost per token at time of use"
        decimal total_cost "Total generation cost"
        timestamp started_at "Generation start time"
        timestamp ended_at "Generation end time"
    }

    teams ||--o{ generation_usage : "team_db_id"
    agents ||--o{ generation_usage : "agent_db_id"
    users ||--o{ generation_usage : "user_db_id"
    model_prices ||--o{ generation_usage : "price_id"
```

### Key Benefits

1. **Generation-level tracking**: Each AI generation gets its own record with `generation_id`
2. **Price versioning**: Tracks which price was used for each generation
3. **Simplified structure**: Two focused tables instead of complex relationships

This maintains the necessary price management while simplifying the usage tracking to generation-level granularity.

rindrics:
Thank you for your feedback. I apologize if [my response to the question about "what granularity we should record"](https://github.com/giselles-ai/giselle/pull/960#issuecomment-2914743247) was misleading. After careful consideration, I believe our current implementation remains the most appropriate for our needs. Let me explain why.

Looking at your proposed changes, I now understand the intent behind the granularity question.
What we ultimately need is "team-based model usage costs (including API calls and image generations in the future) for Stripe billing periods."
When I mentioned that "we need token counts per generation," I didn't mean per generation ID, but rather "per timestamp."
We need timestamps because model pricing might change over time.
While having a reference to generation IDs could be convenient, we don't currently need to investigate usage per specific generation, so our focus remains on "usage" itself.
Therefore, I believe `model_usages` is still a more appropriate table name than generation_usage.

Regarding the "unnecessarily complex" comment, we need to support usage metrics beyond token types (such as API call counts and image metrics) in the future.
This is why we've separated `modelUsageItems` into the separated table and included the `usageMetric` column.

@toyamarinyon

toyamarinyon:
@Rindrics If usage per Generation can be measured with the current table structure and it can be aggregated monthly, that would be good. I made this proposal because I thought it might not be possible.

apps/studio.giselles.ai/drizzle/schema.ts
rindrics: Yes, I'm sorry.
I removed these unused imports in https://github.com/giselles-ai/giselle/pull/960/commits/f1cf29dd2077fea13945ec0498c4d11141dde488 🙇

packages/language-model/src/models.ts
rindrics: After https://github.com/giselles-ai/giselle/pull/960#issuecomment-2915282825, I believe now we can discuss this @toyamarinyon @shige @satococoa 

If we can discuss about the expression of model keys in this PR, I think the key format of `claude-4-opus-20250514-1` is better than the other formats.

toyamarinyon: @Rindrics I want to know why this change is necessary for achieving this Pull request's objective of 'Add DB schema to save model usage'.

rindrics: I introduced model names as constant to ensure consistency between table column `model` and pricing object stored in `language-models/cost/model-pricing.ts`. @toyamarinyon 

ref: https://github.com/giselles-ai/giselle/pull/960#issuecomment-2911840020
> > > 1\. The combination of `model` and `provider` columns in `model_usage` table directly maps to the keys in our `ModelPriceTable`, allowing us to reference the pricing information.
> > 
> > Do these keys necessarily match those of `ModelPriceTable`?
> > If so, there's no problem, but they could not be read from the existing Table definition.
> 
> Thanks, the consistency between these values are not ensured  at that point.
> I fixed this problem by referring model names using constant in https://github.com/giselles-ai/giselle/pull/960/commits/d030a257b592f19e723fdc4b07644608a00642bc 👍

apps/studio.giselles.ai/drizzle/schema.ts
rindrics: Sorry, this was because of the deletion of the column `model` of `model_usages` table.
I re-added the column because it is necessary to create `pricing_id` introduced by https://github.com/giselles-ai/giselle/pull/960/commits/105e1bbec3743e3f0a710eac7e52aaf21cd3fff8

apps/studio.giselles.ai/migrations/0034_glorious_owl.sql
```diff
@@ -0,0 +1,36 @@
+CREATE TABLE IF NOT EXISTS "model_usage" (
+	"db_id" serial PRIMARY KEY NOT NULL,
+	"team_db_id" integer NOT NULL,
+	"model" text NOT NULL,
+	"provider" text NOT NULL,
+	"agent_db_id" integer NOT NULL,
+	"created_at" timestamp DEFAULT now() NOT NULL
+);
```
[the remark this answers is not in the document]
rindrics: Index added by https://github.com/giselles-ai/giselle/pull/960/commits/f76dc6d6c78432eda7728b5576ce62e2d7b4af2b

apps/studio.giselles.ai/migrations/0034_glorious_owl.sql
```diff
@@ -0,0 +1,36 @@
...
+--> statement-breakpoint
+CREATE TABLE IF NOT EXISTS "model_usage_items" (
+	"db_id" serial PRIMARY KEY NOT NULL,
+	"model_usage_db_id" integer NOT NULL,
+	"usage_metric" text NOT NULL,
+	"amount" integer NOT NULL,
+	"unit" text NOT NULL,
+	"ended_at" timestamp NOT NULL,
+	"created_at" timestamp DEFAULT now() NOT NULL
+);
```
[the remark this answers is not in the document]
rindrics: I'll ignore this because index added by https://github.com/giselles-ai/giselle/pull/960/commits/f76dc6d6c78432eda7728b5576ce62e2d7b4af2b

packages/language-model/src/costs/model-prices.ts
```diff
@@ -1,322 +1,284 @@
...
+				output: {
+					costPerMegaToken: 4.4,
 				},
 			},
-		],
-	},
-};
+		},
+	],
+});
```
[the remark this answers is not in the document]
rindrics: The proposed changes below contain hallucination-induced errors and should be disregarded:
```diff
 [OPENAI_MODELS.GPT4O]: [
  {
    validFrom: "2025-05-12T00:00:00Z",
    price: {
-     input: { costPerMegaToken: 2.5 },
-     output: { costPerMegaToken: 10.0 },
+     input: { costPerMegaToken: 3.75 },
+     output: { costPerMegaToken: 15.0 },
    },
  },
 ],
 
 [OPENAI_MODELS.GPT4O_MINI]: [
  {
    validFrom: "2025-05-12T00:00:00Z",
    price: {
-     input: { costPerMegaToken: 0.15 },
-     output: { costPerMegaToken: 0.6 },
+     // TODO: confirm current GPT-4o-mini input/output rates
    },
  },
 ],
```

OpenAI pricing table: https://platform.openai.com/docs/pricing#latest-models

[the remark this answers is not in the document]
rindrics: We’re still figuring out the direction for this file [in another conversation](https://github.com/giselles-ai/giselle/pull/960#discussion_r2110614806), so I’ll mark this resolved for now 🙏

packages/language-model/src/models.ts
toyamarinyon: @Rindrics I think it might be over-engineering this. Instead of creating constant objects for model names, we could follow a simpler approach like Vercel AI does:

https://github.com/vercel/ai/blob/2669f00b8e9acf8352bd07d930fbd181e5219500/packages/openai/src/openai-chat-settings.ts#L1-L35

```typescript
// Simply use string literal types
export type OpenAIModel = 
  | 'gpt-4o'
  | 'gpt-4o-mini'
  | 'gpt-4.1'
  | 'gpt-4.1-mini'
  | 'o3'
  | 'o3-mini'
  // ...

// Use the actual model names directly in the pricing table
export const openAiTokenPricing = defineModelPriceTable({
  'gpt-4o': [{ validFrom: "2025-05-12T00:00:00Z", price: { ... } }],
  'gpt-4o-mini': [{ validFrom: "2025-05-12T00:00:00Z", price: { ... } }],
  // ...
});

// Database schema can use the same type
export const modelUsage = pgTable("model_usage", {
  model: text("model").$type<OpenAIModel>().notNull(),
  // ...
});

rindrics:
Thank you @toyamarinyon 

> **Current Issues:**
- > ❌ Can't calculate costs using SQL queries alone - requires application logic
  - true. I understand this
- > ❌ Price changes require code deployment - business configuration shouldn't need developer intervention  
  - true. This is an important suggestion
- > ❌ Historical pricing is hard to manage - difficult to recalculate costs for past periods
  - false. `priceId` introduced by https://github.com/giselles-ai/giselle/pull/960/commits/105e1bbec3743e3f0a710eac7e52aaf21cd3fff8 enables us to query usage during given period
- > ❌ Non-technical team members can't update pricing - requires code changes
  - true. This is an important suggestion

Although we discussed current design before, would you give me your opinion at this point? 🙏  @satococoa

satococoa:
@toyamarinyon cc/ @Rindrics @shige 

I believe storing model pricing data in our codebase is a reasonable approach.

- Pros:
  - Eliminates the need for database pricing management interfaces or scripts
    - I want to avoid manual database operations
  - Price changes are tracked in git history
  - Enables atomic deployment of new models with pricing
  - Pricing changes require deployment (which I consider an advantage)
- Cons:
  - SQL-based calculations become impossible

satococoa:
Another idea 💡 

- Having pricing table as a seed data for DB.
- automatically apply the data with deployment.

rindrics:
Because the code for billing needs to be carefully implemented, it's better for us to discuss the architecture and come back with another PR.

Thank you for your reviews! @shige @satococoa @toyamarinyon

PR #967, opened 2025-05-27
Fix html-only test in firecrawl

shige:
## Summary
- ensure `html` results have length in `firecrawl.test`

## Testing
- `npx turbo test --cache=local:rw` *(fails: getaddrinfo ENOTFOUND registry.npmjs.org)*

## Summary by CodeRabbit

- **Tests**
  - Improved test coverage by ensuring that scraped HTML content is non-empty when using the "html" option.


shige:
Thank you! 🚀

PR #976, opened 2025-05-28
Refactor GitHub badge rendering

toyamarinyon:
## Summary
- centralize GitHub node status logic in `GitHubNodeInfo`
- update node component to use the new helper

## Testing
- `turbo test --cache=local:rw`
- `turbo check-types --cache=local:rw`


shige (approved):
Thank you! ♻️

PR #978, opened 2025-05-28
Add mock webhook event utility

toyamarinyon:
## Summary
- add a helper for generating minimal GitHub webhook events
- re-export the helper from package entry

## Testing
- `npx turbo build --filter @giselle-sdk/github-tool --cache=local:rw`
- `npx turbo check-types --filter @giselle-sdk/github-tool --cache=local:rw`
- `npx turbo test --filter @giselle-sdk/github-tool --cache=local:rw` *(fails: Missing tasks in project)*


toyamarinyon:
It wasn't needed right now, so I will close it.

PR #987, opened 2025-05-28
Refactor properties panel utilities

toyamarinyon:
## Summary
- add `filterNodeItems` helper to dedupe node filtering logic
- create reusable `ConnectionListRoot` and `ConnectionListItem`
- update input panels to use new list components
- export new components in `ui/index.ts`
- use common filter helper in existing utilities

## Testing
- `npx turbo run check-types --filter=@giselle-internal/workflow-designer-ui --cache=local:rw`
- `npx turbo run test --filter=@giselle-internal/workflow-designer-ui --cache=local:rw` *(fails: Missing tasks in project)*


toyamarinyon:
outdated. I'll try later

PR #997, opened 2025-05-29
Fix regex for HTML title parsing

shige:
## Summary
- avoid using ES2018-only regex flag in self-made scraper

## Testing
- `pnpm -F @giselle-sdk/web-search run check-types`
- `pnpm -F @giselle-sdk/web-search test`

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved compatibility when extracting HTML title tags, ensuring more reliable title detection across different environments.
  - Enhanced validation of webpage title extraction to confirm accurate results during content retrieval.

### DIFF
--- a/packages/web-search/src/self-made.ts
+++ b/packages/web-search/src/self-made.ts
@@ -49,7 +49,7 @@ export async function scrapeUrl(
 	}
 	const html = await res.text();
 	// Extract title from HTML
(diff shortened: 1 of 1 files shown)


shige:
@toyamarinyon 
Thank you!
I added a test case first and confirmed that the test passed before and after the correction. ✅

PR #1046, opened 2025-06-04
Add action result panel

toyamarinyon:
## Summary
- show GitHub action results
- display results using new generation panel

> [!NOTE]
> It's displayed as-is without formatting, so it's hard to read, but we plan to improve this in the future. We also want to display the cause of errors when they occur.

image

## Testing
- `npx turbo build --filter '@giselle-sdk/*' --filter giselle-sdk --cache=local:rw`
- `npx turbo check-types --cache=local:rw`
- `npx turbo test --cache=local:rw`

------
https://chatgpt.com/codex/tasks/task_e_683fffc600ac832f959032b615a04181

## Summary by CodeRabbit

- **New Features**
  - Introduced a panel to display generation results for action nodes, including status and generated text outputs.
  - Added a vertically resizable split view when configuring GitHub action nodes, allowing users to view both configuration and generation results side by side.

- **Improvements**
  - Enhanced interaction by enabling action runs to be triggered from within the properties panel.

### DIFF
--- a/internal-packages/workflow-designer-ui/src/editor/properties-panel/action-node-properties-panel/generation-panel.tsx
+++ b/internal-packages/workflow-designer-ui/src/editor/properties-panel/action-node-properties-panel/generation-panel.tsx
@@ -45,10 +45,8 @@ function Empty({ onRun }: { onRun?: () => void }) {
 
 export function GenerationPanel({
 	node,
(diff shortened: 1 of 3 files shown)


shige (approved):
Thank you! 📄

PR #1052, opened 2025-06-05
New Run Button(wip & feature flag)

toyamarinyon:


shige:
> ### Bug: Header Component Incorrectly Hides Elements
> When the `runV3` feature flag is enabled, the Header component incorrectly hides the `UserPresence`, `ShareButton` (when `shareFeatureFlag` is true), and the element passed via the `action` prop. These elements should remain visible regardless of the `runV3` flag, as they provide independent functionality like collaboration, sharing, and custom actions.

@toyamarinyon 

As Cursor commented, the existing buttons were no longer displayed.

image

shige:
> As Cursor commented, the existing buttons were no longer displayed.

Sorry, I made a mistake. @toyamarinyon 

I just hadn't installed the Trigger Node. 🙏

PR #1063, opened 2025-06-06
Bug Fix: “Start Manual Flow” dialog always shows the same Node

shige:
This PR fixes an issue where clicking the “Run” button and selecting a trigger node would always open the “Start Manual Flow” dialog for the same node, regardless of which node was selected.

**Cause:**
The bug was due to a single dialog open state (`isDialogOpen`) being shared across all trigger nodes in the dropdown. As a result, opening the dialog for any node would always show the same node.

**Fix:**
The dialog open state is now tracked per node using the node’s ID. This ensures that only the dialog for the selected node is open, and the correct node is shown in the dialog.

**How to test:**
- Click the “Run” button.
- Select different trigger nodes from the dropdown.
- The “Start Manual Flow” dialog should now display the correct node each time.

### DIFF
--- a/internal-packages/workflow-designer-ui/src/header/run-button/run-button.tsx
+++ b/internal-packages/workflow-designer-ui/src/header/run-button/run-button.tsx
@@ -45,7 +45,7 @@ export function RunButton() {
 	const { data } = useWorkflowDesigner();
 	const { startFlow } = useFlowController();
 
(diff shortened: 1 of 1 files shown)


shige:
I did my OATs! ✅ 

https://github.com/user-attachments/assets/c2cb6697-d377-43b3-a2fb-b771d2e86ebc

shige:
Thank you! 🚀

PR #1107, opened 2025-06-11
Refactor GitHub search tools

shige:
## Summary
- update GitHub tools to use `@octokit/rest`
- call `octokit.search.issuesAndPullRequests` for searching issues and PRs

## Testing
- `pnpm biome check --write packages/github-tool/src/tools.ts`
- `npx turbo run build --filter @giselle-sdk/github-tool --cache=local:rw` *(fails: cannot find module '@octokit/rest')*
- `npx turbo run check-types --filter @giselle-sdk/github-tool --cache=local:rw` *(fails: cannot find module '@octokit/rest')*

------
https://chatgpt.com/codex/tasks/task_e_6849817c0b948325a3b81e809fe6c439

### DIFF
--- a/packages/github-tool/src/tools.ts
+++ b/packages/github-tool/src/tools.ts
@@ -1,4 +1,4 @@
-import type { Octokit } from "@octokit/core";
+import type { Octokit } from "@octokit/rest";
 import { tool } from "ai";
 import { z } from "zod";
 
@@ -1088,7 +1088,7 @@ export function githubTools(octokit: Octokit) {
 			execute: async (params) => {
 				const { order, page, perPage, q, sort } = params;
 
-				const response = await octokit.request("GET /search/issues", {
+				const response = await octokit.search.issuesAndPullRequests({
 					q,
 					sort,
 					order,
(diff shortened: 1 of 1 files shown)


shige:
### Background

The current implementation relies on the low-level `octokit.request` API.

Because of that, we would have to constantly monitor GitHub’s API docs and deprecation logs to notice when an endpoint becomes outdated.

For example, when Giselle calls `searchIssues` or `searchPullRequests`, we now see this warning (first noticed while implementing `searchPullRequests`):

```
[@octokit/request] "GET https://api.github.com/search/issues?q=xxx" is deprecated.
It is scheduled to be removed on Thu, 04 Sep 2025 00:00:00 GMT.
See https://github.blog/changelog/2025-03-06-github-issues-projects-api-support-for-issues-advanced-search-and-more/
```

To stay aligned with external API changes, I refactored part of the code to use the higher-level methods provided by **`@octokit/rest`** instead of the lower-level `octokit/core`. As long as we keep our packages up to date, we will automatically pick up GitHub API updates.

In this PR I refactored only the two calls most likely to break soon—`searchIssues` and `searchPullRequests`. We could migrate the rest as well, but `@octokit/rest` still exposes `octokit.request`, so the remaining calls can stay as they are for now. Directly calling the REST API through `octokit.request` also has the advantage that we don’t have to wait for `@octokit/rest` to catch up.

shige:
I feel that migrating octokit(authConfig: GitHubAuthConfig) using @octokit/core to @octokit/rest is not worth the effort, so I will close it.

PR #1180, opened 2025-06-19
Refactor workflow utils to arrays

toyamarinyon:
## Summary
- switch workflow utilities from plain objects to arrays
- rename helpers to reflect array-based data
- update workflow building logic for arrays
- adjust tests for new array helpers

## Testing
- `pnpm turbo test --cache=local:rw`

------
https://chatgpt.com/codex/tasks/task_e_68538371cc0c832fadeaa15602a8220b

### DIFF
--- a/packages/workflow-utils/src/build-workflow-from-node.test.ts
+++ b/packages/workflow-utils/src/build-workflow-from-node.test.ts
@@ -5,19 +5,19 @@ import {
 	isTriggerNode,
 } from "@giselle-sdk/data-type";
 import { beforeEach, describe, expect, it, test } from "vitest";
-import { buildWorkflowFromNode } from "./build-workflow-from-node";
+import { buildWorkflowForNode } from "./build-workflow-from-node";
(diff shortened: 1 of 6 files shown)


toyamarinyon:
@shige I'm changing the method for creating workflows. The tests are passing, and I have also checked with several flows, but it would be helpful if you could check with various flows to see if there are any regressions.

shige (changes requested):
Thank you for the refactoring!

It seems that the Text Node output cannot be referenced by the Text Generator Node. Please investigate the cause and fix it!


image

toyamarinyon:
Oh, thanks. I'll fix it

toyamarinyon:
I fixed and added tests!

shige (approved):
Thank you! 👍  

---

📝 Discussion w/ @toyamarinyon 

After discussing this with the Giselle team, we came to the conclusion that the Query Node's prompts would be dynamically generated using a Text Generator Node or similar, which resolved the issue.

PR #1181, opened 2025-06-19
fix: correct task duration calculation in flow runner

toyamarinyon:
## Summary

This PR fixes an issue with task duration calculation in the flow runner. Previously, the task duration was incorrectly calculated using flow-wide timing instead of tracking individual operation durations.

### Changes
- Track individual operation durations during execution
- Calculate total task duration by summing all operation durations
- Rename variables for better clarity:
  - `jobStartedAt` → `flowStart` (more accurate name since it's used for flow-level timing)
  - `totalTasks` → `totalTaskDuration` (clearer about what's being measured)

### Technical Details
The previous implementation had a bug where `totalTasks` was being calculated as `Date.now() - jobStartedAt` for each operation, which would result in increasingly larger values for each subsequent operation. The fix properly tracks the duration of each individual operation and sums them up for the total task duration.

## Test Plan
- [x] Code changes have been reviewed
- [x] No test failures expected as this is a bug fix that corrects the calculation logic
- [ ] Manual testing: Run a flow with multiple operations and verify that task durations are calculated correctly
- [ ] Monitor flow run statistics to ensure duration metrics are accurate

## Impact
This fix ensures accurate tracking of task execution time in flow runs, which is important for performance monitoring and optimization.

### DIFF
--- a/packages/giselle-engine/src/core/flows/run-flow.ts
+++ b/packages/giselle-engine/src/core/flows/run-flow.ts
@@ -153,11 +153,9 @@ async function runJob(args: {
 		},
 	});
 
-	const jobStartedAt = Date.now();
-	let totalTasks = 0;
 	let hasJobError = false;
 
-	await Promise.all(
+	const operationDurations = await Promise.all(
 		job.operations.map(async (operation) => {
 			const generation = createQueuedGeneration({
 				operation,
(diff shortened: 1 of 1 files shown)


shige (approved):
Thank you! ⏲️

PR #1204, opened 2025-06-23
Fix Escape key issue in TourOverlay

shige:
## Summary
Fix Escape key issue in TourOverlay

## Changes
Added a global Escape key listener so pressing Escape exits the overlay and kept keyboard accessibility.

## Testing
1. New App
2. Type Escape key
3. Close the TourOverlay

## Other Information
I fixed this bug in preparation for adding new E2E scenarios.


shige:
@giselles-ai/giselle-developers 
Could you review codes when you have a time?

shige:
Thank you! 🚀

PR #1214, opened 2025-06-24
Mark vector store module as experimental

toyamarinyon:
## Description
This PR renames the `vector-store` module to `experimental_vector-store` to clearly indicate that this feature is in an experimental state and may be subject to changes.

## Changes
- Renamed directory structure from `vector-store` to `experimental_vector-store`
- Updated import paths in affected files to reference the new module name
- Updated exports in core/index.ts to use the new module name
- Fixed all references to maintain consistency across the codebase

## Motivation
By explicitly marking the vector store functionality as experimental, we provide better expectations to users about the stability of this feature. This naming convention helps clarify which parts of the API might be subject to breaking changes in future releases.


shige (approved):
Thank you! ✔️

PR #1263, opened 2025-06-30
Fix workspace cache issue causing data persistence problems

### LINKED ISSUE #1096: Workspace: Auto-save changes before navigating away
## Task description

Implement Autosave Before Navigating Away from Workspace Page.

## Background

Currently, changes made on the Workspace page (e.g., adding or deleting Nodes) are not automatically saved when a user navigates to a different page. This can lead to a loss of work if the user forgets to manually save before leaving.

**Current Behavior (Steps to Reproduce):**
1. Open a Workspace page (e.g., `https://studio.giselles.ai/workspaces/wrks-id`).
2. Add all Nodes one by one.
3. Delete some of the placed Nodes.
4. Navigate to another page (e.g., `https://studio.giselles.ai/apps`).
5. Return to the previous Workspace page (`https://studio.giselles.ai/workspaces/wrks-id`).
6. Observe that the deleted Nodes have reappeared, indicating the changes were not saved.

## TODO (Optional)

- [ ] Implement an autosave function that triggers when a user attempts to navigate away from the Workspace page.
- [ ] If the autosave process has not completed when the user initiates navigation, display a dialog box.
- [ ] This dialog should ask the user whether they want to wait for the save to complete before proceeding with the page transition.

gentamura:
## Summary
- Fixed workspace changes appearing to be lost when navigating away (#1096)
- The issue was caused by cache returning stale data, not save failures

## Changes
- Convert `cacheControlMaxAge` to proper Cache-Control header format in supabase-storage-driver
- Add `bypassingCache` option when loading workspace to ensure fresh data

## Test plan
- [x] Add a node to workspace
- [x] Navigate to another page 
- [x] Return to workspace - node should still be present
- [x] Delete a node from workspace
- [x] Navigate to another page
- [x] Return to workspace - node should remain deleted

## Details
The root cause was that while saves were successful, the storage cache was returning outdated data on subsequent loads. This made it appear as if changes were not persisted. The fix ensures:
1. Write operations set `Cache-Control: max-age=0` correctly
2. Read operations bypass cache to always get fresh data

Fixes #1096

___

### **PR Type**
Bug fix

___

### **Description**
- Fix workspace cache issue causing data persistence problems

- Convert `cacheControlMaxAge` to proper Cache-Control header format

- Add `bypassingCache` option to ensure fresh workspace data

- Resolve issue where workspace changes appeared lost after navigation

___

### **Changes diagram**

```mermaid
flowchart LR
  A["Storage Driver"] --> B["Fix Cache-Control Header"]
  C["Workspace Utils"] --> D["Add Bypass Cache Option"]
  B --> E["Proper Cache Headers"]
  D --> F["Fresh Data Loading"]
  E --> G["Workspace Changes Persist"]
  F --> G
```

___

### **Changes walkthrough** 📝
<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Bug fix</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>supabase-storage-driver.ts</strong><dd><code>Fix Cache-Control header format in storage driver</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

apps/studio.giselles.ai/supabase-storage-driver.ts

<li>Convert <code>cacheControlMaxAge</code> to proper <code>Cache-Control: max-age=X</code> header <br>format<br> <li> Add conditional check for undefined <code>cacheControlMaxAge</code> values

</details>

  </td>
  <td><a href="https://github.com/giselles-ai/giselle/pull/1263/files#diff-13074662ed6a4fe77407e78cb85916f9b72550ea035735641d48516c164c4da3">+4/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td>
    <details>
      <summary><strong>utils.ts</strong><dd><code>Add cache bypass option for workspace loading</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

packages/giselle-engine/src/core/workspaces/utils.ts

<li>Add <code>bypassingCache: true</code> option to <code>getWorkspace</code> function<br> <li> Update comment to reference Supabase instead of Vercel Blob

</details>

  </td>
  <td><a href="https://github.com/giselles-ai/giselle/pull/1263/files#diff-e11c7dcfd0424235ec77ad46ff5727c61347b8c7ec2a601d9bde9722b867a411">+4/-2</a>&nbsp; &nbsp; &nbsp; </td>

</tr>
</table></td></tr></tr></tbody></table>

___

> <details> <summary>  Need help?</summary><li>Type <code>/help how to ...</code> in the comments thread for any questions about Qodo Merge usage.</li><li>Check out the <a href="https://qodo-merge-docs.qodo.ai/usage-guide/">documentation</a> for more information.</li></details>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of cache control settings to ensure compatibility and correct formatting when storing items.
  * Workspace retrieval now explicitly bypasses cache to provide the most up-to-date data.

* **Documentation**
  * Updated comments to clarify cache control behavior for Supabase storage.

### DIFF
--- a/apps/playground/supabase-storage-driver.ts
+++ b/apps/playground/supabase-storage-driver.ts
@@ -113,7 +113,7 @@ export default defineDriver((options: SupabaseStorageDriverOptions) => {
 		},
 
 		async setItem(key, value, opts) {
-			const cacheControl = opts?.cacheControlMaxAge;
+			const cacheControl = opts?.cacheControl;
 			const contentType = opts?.contentType;
 
 			const { data, error } = await supabase.storage
--- a/apps/studio.giselles.ai/supabase-storage-driver.ts
+++ b/apps/studio.giselles.ai/supabase-storage-driver.ts
@@ -124,10 +124,7 @@ export default defineDriver((options: SupabaseStorageDriverOptions) => {
 		},
 
 		async setItem(key, value, opts) {
-			const cacheControl =
-				opts?.cacheControlMaxAge !== undefined
-					? `max-age=${opts.cacheControlMaxAge}`
-					: undefined;
+			const cacheControl = opts?.cacheControl;
 			const contentType = opts?.contentType;
 
 			const { data, error } = await supabase.storage
--- a/apps/studio.giselles.ai/supabase-storage-driver.ts
+++ b/apps/studio.giselles.ai/supabase-storage-driver.ts
@@ -124,7 +124,10 @@ export default defineDriver((options: SupabaseStorageDriverOptions) => {
 		},
 
 		async setItem(key, value, opts) {
-			const cacheControl = opts?.cacheControlMaxAge;
+			const cacheControl =
+				opts?.cacheControlMaxAge !== undefined
+					? `max-age=${opts.cacheControlMaxAge}`
+					: undefined;
 			const contentType = opts?.contentType;
 
 			const { data, error } = await supabase.storage
--- a/packages/giselle-engine/src/core/workspaces/utils.ts
+++ b/packages/giselle-engine/src/core/workspaces/utils.ts
@@ -16,7 +16,7 @@ export async function setWorkspace({
 	workspace: Workspace;
 }) {
 	await storage.setItem(workspacePath(workspaceId), workspace, {
-		// Disable caching by setting cacheControlMaxAge to 0 for Vercel Blob storage
+		// Disable caching by setting cacheControlMaxAge to 0 for Supabase storage
 		cacheControlMaxAge: 0,
 	});
 }
@@ -28,7 +28,9 @@ export async function getWorkspace({
 	storage: Storage;
 	workspaceId: WorkspaceId;
 }) {
-	const result = await storage.getItem(workspacePath(workspaceId));
+	const result = await storage.getItem(workspacePath(workspaceId), {
+		bypassingCache: true,
+	});
 	const workspace = parseAndMod(
 		Workspace,
 		result,
--- a/packages/giselle-engine/src/core/workspaces/utils.ts
+++ b/packages/giselle-engine/src/core/workspaces/utils.ts
@@ -16,8 +16,8 @@ export async function setWorkspace({
 	workspace: Workspace;
 }) {
 	await storage.setItem(workspacePath(workspaceId), workspace, {
-		// Disable caching by setting cacheControlMaxAge to 0 for Supabase storage
-		cacheControlMaxAge: 0,
+		// Disable caching by setting cacheControl to 0 for Supabase storage
+		cacheControl: 0,
 	});
 }
 


gentamura:
@shige @toyamarinyon Could you please review this code? :pray:

shige (approved):
Thank you! 🦺

toyamarinyon (commented):
I thought this might not be able to solve the problem. Rather, since this issue occurs when navigating to a different screen during the interval before saving, I think it's necessary to handle the navigation and perform an immediate save at that time.

gentamura:
> I thought this might not be able to solve the problem. Rather, since this issue occurs when navigating to a different screen during the interval before saving, I think it's necessary to handle the navigation and perform an immediate save at that time.

@toyamarinyon Thank you for your comment! I initially had the same thought and started investigating the save mechanism.

What I discovered was that the save operation to Supabase storage actually completes instantly. The issue occurred when navigating away and then returning to the playground - the old workspace.json was being returned, which made it appear as if nodes weren't added or deleted.

I couldn't definitively determine whether the caching was happening at the browser level or storage level, so I addressed it by implementing cache busting and adjusting the storage max-age settings. This approach seems to have resolved the issue.

packages/giselle-engine/src/core/workspaces/utils.ts
```diff
@@ -16,7 +16,7 @@ export async function setWorkspace({
 	workspace: Workspace;
 }) {
 	await storage.setItem(workspacePath(workspaceId), workspace, {
-		// Disable caching by setting cacheControlMaxAge to 0 for Vercel Blob storage
+		// Disable caching by setting cacheControlMaxAge to 0 for Supabase storage
```
toyamarinyon: It was recognized that the Vercel Blob option is not supported by Superbase Storage.

toyamarinyon (commented):
I see, I understand that. I've added comments to the code.

apps/studio.giselles.ai/supabase-storage-driver.ts
```diff
@@ -124,7 +124,10 @@ export default defineDriver((options: SupabaseStorageDriverOptions) => {
 		},
 
 		async setItem(key, value, opts) {
-			const cacheControl = opts?.cacheControlMaxAge;
+			const cacheControl =
+				opts?.cacheControlMaxAge !== undefined
+					? `max-age=${opts.cacheControlMaxAge}`
+					: undefined;
```
toyamarinyon: I couldn't find anything about cacheControlMaxAge in the Superbase Storage documentation, is it somewhere?

gentamura: @toyamarinyon Thanks for your review. I checked both the documentation and the type definitions for Supabase Storage, and it looks like the value passed to `cacheControl` should be a string representing the number of seconds (not a full `max-age=` directive):

[https://supabase.com/docs/reference/javascript/storage-from-upload](https://supabase.com/docs/reference/javascript/storage-from-upload)

```ts
// Excerpt from:
// giselle/node_modules/.pnpm/@supabase+storage-js@2.7.1/node_modules/@supabase/storage-js/src/lib/types.ts:45-50
export interface FileOptions {
  /**
   * The number of seconds the asset is cached in the browser and in the Supabase CDN. This is set in the `Cache-Control: max-age=<seconds>` header. Defaults to 3600 seconds.
   */
  cacheControl?: string
}
```

I’ve also updated the property name in the `opts` object to align with Supabase Storage’s option naming for consistency.

packages/giselle-engine/src/core/workspaces/utils.ts
gentamura: ref: https://github.com/giselles-ai/giselle/pull/1263#discussion_r2181271872

apps/playground/supabase-storage-driver.ts
```diff
@@ -113,7 +113,7 @@ export default defineDriver((options: SupabaseStorageDriverOptions) => {
 		},
 
 		async setItem(key, value, opts) {
-			const cacheControl = opts?.cacheControlMaxAge;
+			const cacheControl = opts?.cacheControl;
```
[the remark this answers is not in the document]
gentamura: Thanks for the suggestion. However, the current approach is more appropriate here since `opts` follows the unstorage framework's `TransactionOptions` pattern (`Record<string, any>`), which is designed for flexibility. Different methods use different option properties (`cacheControl`, `contentType`, `bypassingCache`, etc.), so creating specific interfaces would add maintenance overhead and go against the framework's flexible design.

shige:
Thank you! 🚀

PR #1305, opened 2025-07-02
Fix path validation to allow Next.js catch-all routes

satococoa:
## Summary
- Fixed overly strict path validation that was rejecting valid Next.js catch-all route patterns like `[...next-giselle]`
- Changed validation to check for actual directory traversal segments (`..`) instead of substring matching
- Maintains security while allowing valid Next.js patterns

## Test plan
- [ ] Verify that Next.js catch-all routes like `apps/playground/app/api/giselle/[...next-giselle]/route.ts` are now properly ingested
- [ ] Confirm that actual path traversal attempts like `../../../etc/passwd` are still blocked
- [ ] Test with various valid and invalid path patterns to ensure security is maintained

### DIFF
--- a/packages/github-tool/src/blob-download-loader.ts
+++ b/packages/github-tool/src/blob-download-loader.ts
@@ -94,9 +94,9 @@ export function createGitHubBlobDownloadLoader(
 
 	function isValidPath(path: string): boolean {
 		// Prevent path traversal attacks
-		return (
-			!path.includes("..") && !path.startsWith("/") && !path.includes("\0")
-		);
+		const segments = path.split(/[/\\]/);
+		const hasTraversal = segments.some((segment) => segment === "..");
+		return !hasTraversal && !path.startsWith("/") && !path.includes("\0");
 	}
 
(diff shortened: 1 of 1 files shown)


satococoa:
> Path traversal validation:The change from substring matching to segment-based validation is generally more secure, but edge cases need verification. Potential concerns include: 1) Empty segments from consecutive separators like "path//../../file" might not be properly handled, 2) Mixed path separators could create bypass opportunities, 3) URL-encoded or Unicode variations of ".." might not be detected. The regex /[/\\]/ correctly handles both forward and backslashes, but comprehensive testing with malicious path patterns is essential.

The current implementation handles the basic cases:
- Empty segments are handled correctly
- Mixed separators are unified by the regex
- No URL decoding prevents encoded traversal attempts

For a general-purpose path validator, additional considerations would be needed.
But it works well for GitHub tarball extraction where:
-  File paths come from a git repository
-  The secondary check in loadDocument() catches any edge cases
-  Symbolic links are already filtered out

shige (approved):
Thank you! 👍

PR #1330, opened 2025-07-02
Upgrade to Biome v2

shige:
## Summary
Upgrade to Biome v2

## Changes
https://biomejs.dev/guides/upgrade-to-biome-v2/
    
```
pnpm biome migrate --write
pnpm biome check --write .
```

fix: resolve Biome v2 linting errors and warnings
    
- Remove unused variables and imports across the codebase
- Fix async functions without await expressions
- Add proper biome-ignore comments for accessibility warnings
- Fix suppression comments with proper explanations
- Remove duplicate type aliases
- Fix unused function parameters
- Replace empty fragments with null returns

## Testing
pnpm ci .

## Other Information
🤖 Generated with [Claude Code](https://claude.ai/code)
    
Co-Authored-By: Claude <noreply@anthropic.com>

___

### **PR Type**
Enhancement, Bug fix

___

### **Description**
• Upgrade Biome linter from v1.9.4 to v2.0.6 across all packages
• Update Biome configuration files to use v2 syntax (`includes` instead of `ignore`, new assist actions)
• Fix numerous linting violations introduced by stricter Biome v2 rules:
  - Remove unused imports, variables, and function parameters across the codebase
  - Rename unused variables with underscore prefix to indicate intentional non-usage
  - Remove unnecessary `async` keywords from functions without await expressions
  - Replace empty React fragments with `null` returns where appropriate
  - Add proper accessibility attributes and biome-ignore comments for accessibility warnings
• Reorganize imports consistently throughout the codebase (external packages before internal modules)
• Alphabetically sort exports and imports for better maintainability
• Clean up function parameter formatting and JSX structure
• Remove duplicate type aliases and consolidate related imports

___

### **Changes walkthrough** 📝
<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Formatting</strong></td><td><details><summary>63 files</summary><table>
<tr>
  <td>
    <details>
      <summary><strong>index.ts</strong><dd><code>Clean up imports and reorganize exports in core engine</code>&nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

packages/giselle-engine/src/core/index.ts

• Remove unused import <code>calculateDisplayCost</code> from language-model <br>package<br> • Reorder imports and exports alphabetically for better <br>organization<br> • Simplify function parameter formatting by removing <br>unnecessary line breaks

</details>

  </td>
[...]

PR #1567, opened 2025-07-29
fix(github-tool): switch to files API to avoid 300 file limit

satococoa:
## Summary
- Switch from diff endpoint to files endpoint to handle PRs with more than 300 files
- Implement proper pagination using Link header parsing
- Fix issue where large PRs would fail with 422 error

## Problem
The previous implementation using `/repos/{owner}/{repo}/pulls/{pull_number}` with `accept: application/vnd.github.v3.diff` would fail with a 422 error when a PR had more than 300 files.

## Solution
Use `/repos/{owner}/{repo}/pulls/{pull_number}/files` endpoint with pagination:
- Fetch files in batches of 100
- Parse Link header for accurate pagination end detection
- Skip files without patch data (binary files, etc.)

## Trade-offs
- **Pros**: Handles PRs of any size, more flexible filtering options
- **Cons**: Multiple API calls required (but cached), slightly increased rate limit usage

## Test plan
- [x] Existing tests pass
- [ ] Test with PRs having >300 files (e.g., giselles-ai/giselle#1330)
- [ ] Test with normal PRs to ensure no regression

### DIFF
--- a/packages/github-tool/src/document-loaders/pull-requests/fetchers.ts
+++ b/packages/github-tool/src/document-loaders/pull-requests/fetchers.ts
@@ -19,36 +19,44 @@ export type DiffFetchContext = {
 	repo: string;
 };
 
-// Retrieve the diff for the entire pull request using the REST API and split it by file
 export async function fetchDiffs(
 	ctx: DiffFetchContext,
 	prNumber: number,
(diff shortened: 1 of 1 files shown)


shige (approved):
Thank you! 🆙

PR #1578, opened 2025-07-31
fix-imports-and-update-tests

toyamarinyon:
- Replaced workspace1 fixture with detailed groupNodesFixture for tests  
- Corrected import path in group-nodes.test.ts to fix test errors  
- Removed deprecated test workspace and related tests  
- Fixed import path for sliceGraphFromNode in use-node-groups.ts to resolve module path error


shige (approved):
Thank you! 🧪

PR #1644, opened 2025-08-12
feat(log): add basic GET error route that logs request path and

toyamarinyon:


toyamarinyon:
Time has passed, so I'll think about it again.

PR #1653, opened 2025-08-13
refactor(stage): remove unused sidebar and account fetch, inline form layout and

toyamarinyon:
- Removed unused sidebar component
- Removed unnecessary account fetch logic
- Inlined form layout CSS/markup to simplify structure


shige (approved):
Thank you! ♻️

PR #1688, opened 2025-08-19
refactor: move embedding profile utilities to data-type package

satococoa:
## Summary
- Move `EMBEDDING_PROFILES` constant to data-type package for better reusability
- Move `isEmbeddingProfileId` utility function to data-type package
- Clean up exports and improve module organization

## Why
- These values need to be referenced from the UI, so they must be importable from frontend code
- The data-type package is accessible from both frontend and backend, making it the ideal location


shige (approved):
Thank you! ♻️

PR #1713, opened 2025-08-20
fix(layout): avoid redundant await by returning act directly and pass it to Sidebar

toyamarinyon:
- Return `` directly instead of awaiting it redundantly.
- Pass the returned `act` to `Sidebar` to avoid duplicate awaiting and streamline flow.


toyamarinyon:
i'll fix build error

shige (approved):
Thank you! ⚡

PR #1720, opened 2025-08-21
refactor(ui): unused deterministic color logic and constants

toyamarinyon:
This pull request removes the deterministic color generation logic from the `AgentCard` component. The code for assigning a unique color to each agent based on their ID has been deleted, simplifying the component.

Code simplification:

* Removed the `colors` array, the `stringToHash` function, and the `getDeterministicColor` function, which were previously used to generate a unique color for each agent based on their ID.
* Removed the usage of the `_color` variable, eliminating the color assignment from the `AgentCard` component.


shige (approved):
Thank you! ♻️

PR #1758, opened 2025-08-28
fix(scope): default filter to "all" when none provided to align UX

toyamarinyon:
- Default the scope filter to "all" when no filter is provided, aligning behavior with expected UX.


shige (approved):
Thank you! 🗺️

PR #1767, opened 2025-08-29
chore: remove multiEmbeddingFlag and enable multi-embedding by default

satococoa:
- Drop conditional logic for multi-embedding feature flags across Studio and internal UI
- Simplify vector store pages and dialogs to always show multi-embedding flows
- Remove legacy repository-item-v2 component and related branching
- Streamline feature flag context to exclude multiEmbedding flag
- Update workspace providers to assume multi-profile support


shige (approved):
Thank you! 🚀

PR #1773, opened 2025-08-29
feat-add-logger-types-and-fix-log

toyamarinyon:
- Add logger types, a noop logger, and wire logger through the engine
- Add Pino integration and align Giselle logger with Pino
- Remove `child` from logger type and tighten noop logger signature
- Fix log message: change debug text from "hello" to "generating text"


shige (approved):
Thank you! 🔍

PR #1788, opened 2025-09-02
feat-workflow-reactflow-store

toyamarinyon:
- Add WorkflowStore React context and replace previous designer hook with the new context
- Introduce lightweight React Flow editor, ReactFlow node canvas, node components and debug views for UI state rendering
- Persist node position and selection from React Flow moves; validate node ids before applying changes
- Replace input/output id maps with connection maps and derive/memoize edges from connections
- Optimize rendering: selective Zustand store selectors, concise selectors, selective rendering to avoid full re-renders, and caching/gradient overlay for nodes
- Reorganize editor/store exports/imports and fix relative import paths to prevent errors
- Add node click handling, inspected node state, properties panel, and clear-inspect on canvas clicks
- Improve store robustness: logger, exhaustive change handling, preserve node list and guard removals in onNodesChange
- Fix UI issues: simplify node class lists, reorder imports/formatting to satisfy linter
- Add feature flag (newEditorFlag) and gate workspace page; upgrade pino dependency to 9.0 and clean up specifiers


toyamarinyon:
There were some extra differences included, so I'll organize them.

toyamarinyon:
@shige I'm currently implementing a new Editor that suppresses canvas redrawing. We are testing it on a different page from the existing one, and if the FeatureFlag is not enabled, a 404 page will be displayed, so there will be no impact on existing users.

shige (approved):
Thank you! 🆙

PR #1793, opened 2025-09-02
feat: Add issues.labaled to events

WashizuRyo:

### DIFF
--- a/packages/giselle/src/engine/github/handle-webhook-v2.ts
+++ b/packages/giselle/src/engine/github/handle-webhook-v2.ts
@@ -22,6 +22,7 @@ const events: WebhookEventName[] = [
 	"issues.opened",
 	"issues.closed",
 	"issue_comment.created",
+	"issues.labeled",
 	"pull_request_review_comment.created",
 	"pull_request.opened",
 	"pull_request.ready_for_review",


internal-packages/workflow-designer-ui/src/editor/properties-panel/trigger-node-properties-panel/providers/github-trigger/utils/trigger-configuration.ts
```diff
@@ -30,11 +30,12 @@ interface TriggerConfigResult {
 }
 
 /**
- * Creates a GitHubFlowTriggerEvent based on the event ID and optional callsign
+ * Creates a GitHubFlowTriggerEvent based on the event ID and optional callsign/labels
  */
 export function createTriggerEvent(
 	eventId: GitHubTriggerEventId,
 	callsign?: string,
+	labels?: string[],
```
toyamarinyon: This method requires the user to declare unnecessary variables as undefined, like so:
```ts
createTriggerEvent(eventId, undefined, validLabels);
```
This is unnecessary.
I'd like to be able to specify the necessary variables by passing them as an object, like so:
```ts
createTriggerEvent({ eventId, label }).
```

We want the createTriggerEvent interface to look like this:

```ts
type CreateTriggerEvent = (args: {
    eventId: GitHubTriggerEventId,
    callsign?: string,
    labels?: string[],
}) => GitHubFlowTriggerEvent
```

WashizuRyo: Thanks for the review!
I’ve refactored the code!

shige:
🚀

PR #1818, opened 2025-09-10
fix(ui): render failure details including id and error message for generation

toyamarinyon:
- Render failure details for generation failures
- Include failure ID in the UI
- Display the error message associated with the failure


shige (approved):
Thank you! 🆔

PR #1893, opened 2025-09-30
Restore the type for the unnecessary type example

colinrymer:
Restored a missing return type annotation to createUser function used as an example of unnecessary typing that was removed during a previous edit of the agent definition.

### DIFF
--- a/.claude/agents/ts-coder.md
+++ b/.claude/agents/ts-coder.md
@@ -218,7 +218,7 @@ function formatDate(date: Date) {
 }
 
 // Over-engineered: Explicit types for self-evident returns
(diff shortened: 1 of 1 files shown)


shige (approved):
That's true!

[The previous fix](https://github.com/giselles-ai/giselle/commit/dc7d034bd985fe6e06cde81be66064aaf66113b8#diff-80190795fa2aa69e9f8934df784610b1d3dcfa721f9b1cbc88838d586d599e32L222) has corrected all of the "over-engineered" sample code to the correct sample code.

Thank you for noticing! 💯

PR #1919, opened 2025-10-03
chore(ci): update release workflow to modern checkout, node, pnpm, build and deploy steps

toyamarinyon:
- Update release workflow to use modern actions and steps:
  - Replace checkout step with the latest checkout action
  - Use modern Node setup action
  - Switch to pnpm for package management
  - Add/modernize build step
  - Update deploy step to current recommended approach


shige (approved):
Thank you! 👍 

I think they are currently making modifications to the initial implementation, so I will approve it.

PR #1934, opened 2025-10-06
chore(settings): replace bg-/border-* (white/black) with semantic aliases (Phase 3)

kaochannel154:
…ases in document-vector-store-item


.reports/settings-bg-white-black.txt
```diff
@@ -0,0 +1,30 @@
+  10 apps/studio.giselles.ai/app/(main)/settings/team/vector-stores/document/document-vector-store-item.tsx
```
shige: @kaochannel154 I don't think you need to commit the files in `.reports/`, so could you please delete them and add them to .gitignore?

kaochannel154: Thank you! I’ll get on it right away.

toyamarinyon (approved):

shige (approved):
Thank you! 🚀

PR #1954, opened 2025-10-07
chore(eslint-warn-like): add non-blocking TS/TSX color guard in CI

kaochannel154:
This PR adds a non-blocking code color guard for TS/TSX.

- scripts/lint-colors-code.mjs: warns on
  - Literal colors: hex, rgb/rgba, hsl/hsla, oklch
  - Tailwind direct black/white scales in class strings (bg-/border-/text-*)
- package.json: add pnpm lint:colors:code
- CI: run the guard step (continue-on-error)

Goal: provide visibility of new raw color usages without blocking merges. No runtime behavior changes.


shige:
🚀

PR #2052, opened 2025-10-27
fix(flags): default undefined edge flag to true instead of false

toyamarinyon:
Switch from unstorage to our own implementation of giselle storage.

## Summary by CodeRabbit

* **New Features**
  * Experimental storage functionality is now enabled by default when configuration is not explicitly set.


shige (approved):
Thank you! 🎏

PR #2209, opened 2025-11-18
⚠️ Fix task status override bug - Failed tasks incorrectly marked as completed

toyamarinyon:
## Overview

This PR modifies the task execution completion logic to explicitly set the task status to "completed" after execution. However, this change introduces critical regressions in task failure reporting and duration tracking that need to be addressed.

⚠️ **Breaking Changes**: This PR contains breaking changes that affect task status semantics and may impact existing monitoring/alerting systems.

## Changes

- Modified `executeTask` function in `task-execution-utils` to add an explicit `opts.applyPatches` call that unconditionally sets task status to "completed"
- This change overrides the previous conditional logic that properly set status to either "failed" or "completed" based on execution outcome
- The duration update patch is no longer being applied, potentially breaking task timing metrics

## Testing

**🔴 Critical: Additional testing required before merging**

- [ ] Verify task failure scenarios still properly report "failed" status
- [ ] Confirm duration.wallClock values are correctly persisted
- [ ] Test impact on existing monitoring/alerting systems that depend on task status
- [ ] Validate dashboards and metrics that rely on task duration data

## Review Notes

**⚠️ This PR requires careful review due to the following concerns:**

1. **Regression in failure reporting**: The new code unconditionally marks tasks as "completed", overriding any "failed" status. This breaks error tracking and alerting.

2. **Missing duration updates**: The duration patch is no longer being applied, which may break metrics and performance monitoring.

3. **Architectural concern**: The change appears to duplicate or bypass existing status management logic without clear justification.

**Specific areas needing reviewer attention:**
- Should we preserve the conditional status logic (failed vs completed)?
- How should duration updates be handled alongside the status update?
- Are there integration tests covering these task lifecycle scenarios?

## Related Issues

*Please link any related issues or tickets that provide context for this change.*

---

**Recommendation**: This PR should not be merged in its current state due to the critical regressions it introduces. Consider revising to preserve both proper failure reporting and duration tracking functionality.

### DIFF
--- a/packages/giselle/src/tasks/shared/task-execution-utils.ts
+++ b/packages/giselle/src/tasks/shared/task-execution-utils.ts
@@ -324,7 +324,6 @@ export async function executeTask(
 		patches.status.set(hasError ? "failed" : "completed"),
 		patches.duration.wallClock.set(duration),
 	]);
-	await opts.applyPatches(opts.task.id, [patches.status.set("completed")]);
 
 	await opts.onTaskComplete?.(hasError, duration);
 


toyamarinyon:
I thought there was a problem, but I was mistaken. I discovered it during testing.

PR #2727, opened 2026-02-12
feat: add vercel-ai-sdk.mdc rule to prevent v6 hallucinations

cubaseuser123:
## Summary
I started using the Vercel AI SDK when version 6 was released. I've noticed that when I ask an agent (Cursor, Antigravity, Claude) to write code, it often defaults to version 4 or 5 patterns.

These agents frequently confuse `toolContext` with `experimental_context`, use the wrong `generateObject` syntax, and get `execute` signatures wrong.

To address this, I created a Cursor rule that teaches agents how to avoid these hallucinations. This approach completely stopped the errors for me.

## Related Issue
https://github.com/vercel-labs/agent-skills/issues/133

## Changes
Added `.cursor/rules/vercel-ai-sdk.mdc` — a companion to the existing `update-ai-sdk.mdc` that focuses on writing correct v6 code. It covers 7 hallucination-prone API changes:

1. `toolContext` → `experimental_context`
2. `generateObject` → `Output.object()` (Legacy)
3. `parameters` → `inputSchema`
4. `.strict()` on Zod → `strict: true` on tool
5. `toDataStreamResponse()` → `toUIMessageStreamResponse()`
6. `maxSteps` → `stopWhen: stepCountIs(N)`
7. Raw messages → `UIMessage` + `convertToModelMessages()`

All patterns are sourced from official Vercel AI SDK docs.

## Testing
N/A — documentation-only change, no runtime code affected.

## Other Information
Every code example includes a ✅ CORRECT / ❌ WRONG comparison with links to official docs. A canonical Route Handler reference is included at the end as a quick-copy template.

## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive Vercel AI SDK v6 usage guidelines covering updated patterns for tool calls, input handling, and streaming.
  * Included migration notes for deprecated APIs and recommended replacements to ensure compatibility with the pinned SDK.
  * Added canonical examples and a route handler walkthrough to demonstrate correct, production-ready integration.

### DIFF
--- a/.cursor/rules/vercel-ai-sdk.mdc
+++ b/.cursor/rules/vercel-ai-sdk.mdc
@@ -1,0 +1,242 @@
+---
+description: Vercel AI SDK v6 patterns — prevents common hallucinations of legacy APIs
+globs: *.tsx,*.ts
+alwaysApply: false
+---
+
+# Vercel AI SDK — Correct v6 Patterns
+
+LLMs frequently hallucinate legacy AI SDK APIs. Follow these rules strictly.
+
+## 1. `toolContext` → `experimental_context`
+
+**NEVER** use `toolContext`. It was removed.
+
+```typescript
+// ✅ CORRECT
+import { tool } from 'ai';
+import { z } from 'zod';
+
+const myTool = tool({
+  description: 'Do something with user context',
+  inputSchema: z.object({
+    query: z.string().describe('The search query'),
+  }),
+  execute: async (input, { experimental_context }) => {
+    const ctx = experimental_context as { userId: string };
+    // use ctx.userId
+  },
+});
+
+// Pass context at the call site:
+const result = await generateText({
+  model,
+  tools: { myTool },
+  experimental_context: { userId: '123' },
+  // ...
+});
+```
+
+```typescript
+// ❌ WRONG — v5 pattern, will fail silently
+execute: async (args, { toolContext }) => { ... }
+```
+
+**Source:** https://sdk.vercel.ai/docs/ai-sdk-core/tools-and-tool-calling#context-experimental
+
+---
+
+## 2. `generateObject` is Legacy → use `Output.object()`
+
+`generateObject` and `streamObject` are **deprecated**. Use `generateText` / `streamText` with the `output` property instead.
+
+```typescript
+// ✅ CORRECT
+import { generateText, Output } from 'ai';
+import { z } from 'zod';
+
+const { output } = await generateText({
+  model,
+  output: Output.object({
+    schema: z.object({
+      name: z.string(),
+      ingredients: z.array(z.object({ name: z.string(), amount: z.string() })),
+    }),
+  }),
+  prompt: 'Generate a lasagna recipe.',
+});
+```
+
+```typescript
+// ❌ WRONG — Legacy API
+import { generateObject } from 'ai';
+const { object } = await generateObject({ model, schema, prompt });
+```
+
+**Source:** https://sdk.vercel.ai/docs/ai-sdk-core/generating-structured-data#generateobject-and-streamobject-legacy
+
+---
+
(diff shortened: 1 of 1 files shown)


shige (changes requested):
Thanks for the contribution! Preventing AI SDK hallucinations is a great idea — we've definitely seen agents generate outdated patterns. However, there are a few issues that need to be addressed before we can merge this.

**1. Version mismatch: our project is on AI SDK v5, not v6**

We currently pin `ai: 5.0.101` in `pnpm-workspace.yaml`. Several patterns in this rule are v6-specific and would be **incorrect guidance** for our codebase:

- `generateObject` → `Output.object()`: `generateObject` is still valid in v5
- `maxSteps` → `stopWhen: stepCountIs(N)`: `maxSteps` is the correct option in v5
- `toDataStreamResponse()` → `toUIMessageStreamResponse()`: our codebase actually uses `toUIMessageStream()` (neither of the two options listed)

Could you verify each pattern against v5 and adjust accordingly?

**2. Code style: single quotes → double quotes**

Our project uses Biome with double quotes as the standard. All code examples in the rule use single quotes, which would cause agents to generate code that fails our formatter. Please update all examples to use double quotes.

**3. Overly broad glob scope**

`globs: *.tsx,*.ts` applies this rule to every TypeScript file, including UI components, tests, and database schemas that never touch the AI SDK. Consider narrowing the glob to relevant directories (e.g., `packages/giselle/src/generations/**`) or removing it so the rule is only triggered manually.

**4. Canonical example doesn't match project patterns**

The route handler example uses `model: 'anthropic/claude-sonnet-4.5'` as a raw string, but we use a language model registry pattern. This could steer agents away from our established conventions.

**5. Relationship with existing `update-ai-sdk.mdc`**

We already have `.cursor/rules/update-ai-sdk.mdc`. Could you clarify how these two rules relate? It might make sense to merge them or clearly differentiate their purposes.

cubaseuser123:
Thanks for the detailed review! I went through your codebase, specifically the files generate-content.ts, both Postgres tools, and biome.json, to ensure every pattern matches what you actually use. Here are the changes:

1. Version mismatch, fixed.
I reframed the entire rule from "v6" to "correct patterns for this project." I removed generateObject to Output.object() as a strict rule since generateObject is valid at your pinned version.

One suggestion that I would like to make: Since update-ai-sdk.mdc exists specifically for version upgrades, I added a brief note in Section 7 to mention that generateObject is deprecated upstream in v6. When you upgrade, this rule will already be ready. I can remove this note if you prefer to keep the rule focused only on v5.

2. Double quotes, fixed.
All code examples now use double quotes to match your Biome config (quoteStyle: "double").

3. Globs narrowed, fixed.
I changed from *.tsx,*.ts to:

packages/giselle/src/generations/**/*.ts

4. Canonical example, rewritten.
I rewrote the reference example to match your actual patterns:
createGateway() from @ai-sdk/gateway instead of raw model strings,
toUIMessageStream() instead of toUIMessageStreamResponse(),
smoothStream() with your line-chunking config,
Custom stopWhen pattern alongside stepCountIs (both variants from your generate-content.ts).

5. Relationship with update-ai-sdk.mdc, clarified.
I added a note at the top of the rule:

How this rule differs from update-ai-sdk.mdc: That rule covers upgrading SDK versions (running pnpm outdated, updating the catalog). This rule is about writing correct code against the current pinned SDK version.

shige:
Thank you for taking the time to open this pull request, and we are truly sorry for leaving it unattended for so long.

We are no longer able to actively maintain the project in a way that lets us review external contributions, so we have changed our contribution policy. You can read the details here: https://github.com/giselles-ai/giselle/pull/2941

For this reason, we are very sorry, but we will be closing this pull request. We sincerely appreciate your interest in and contribution to Giselle.

PR #2728, opened 2026-02-12
fix: don’t treat SVG uploads as vision images

### LINKED ISSUE #2714: SVG images incorrectly handled as ImagePart causing "Unsupported MIME type" error
## Description

When an SVG file is uploaded and used in a generation, the system throws an error:

```
UnknownError: Unsupported MIME type: image/svg+xml
```

## Root Cause

In `packages/giselle/src/generations/internal/use-generation-executor.ts:570`, `image/svg+xml` is treated as an `ImagePart` and sent to the LLM provider. However, none of the major LLM providers support SVG format for vision/image input.

## Supported Image Formats by Provider

| Provider | Supported Formats | SVG Support |
|----------|------------------|-------------|
| **OpenAI** | JPEG, PNG, GIF, WebP | Not supported |
| **Anthropic (Claude)** | JPEG, PNG, GIF, WebP | Not supported |
| **Google (Gemini)** | JPEG, PNG, WebP, HEIC, HEIF | Not supported |

### References
- [OpenAI - Images and vision](https://platform.openai.com/docs/guides/images-vision)
- [Anthropic - Vision](https://docs.anthropic.com/en/docs/build-with-claude/vision)
- [Google - Image understanding](https://ai.google.dev/gemini-api/docs/image-understanding)

## Proposed Solutions

1. **Remove SVG from ImagePart handling** - Treat SVG as unsupported and log a warning (simplest)
2. **Convert SVG to raster format** - Server-side conversion to PNG before sending to LLM
3. **Treat SVG as text** - Since SVG is XML, it could potentially be sent as text content

## Current Code

```typescript
switch (fileParameter.type) {
  case "image/jpeg":
  case "image/png":
  case "image/gif":
  case "image/svg+xml":  // <- This should not be treated as ImagePart
    return {
      type: "image",
      image: blob,
    } as ImagePart;
```

KerwinLarrobis:
## Summary
Prevent `image/svg+xml` files from being treated as vision `ImagePart` input.

Instead of sending SVG files to LLM providers (which do not support SVG), this change logs a warning and skips the file.

## Why
LLM vision providers (OpenAI, Anthropic, Google) do not support SVG image input and return:
`Unsupported MIME type: image/svg+xml`.

This prevents SVG files from being passed as ImagePart objects.

## Testing
pnpm -C packages/giselle test

All tests in the giselle package pass.

Closes #2714

## Summary by CodeRabbit

* **New Features**
  * Added WebP image format support for vision input processing.

* **Bug Fixes**
  * SVG images are now properly rejected during vision input processing with a warning message, preventing improper handling of unsupported formats.

### DIFF
--- a/packages/giselle/src/generations/internal/use-generation-executor.ts
+++ b/packages/giselle/src/generations/internal/use-generation-executor.ts
@@ -567,11 +567,19 @@ export async function useGenerationExecutor<T>(args: {
 							case "image/jpeg":
 							case "image/png":
 							case "image/gif":
-							case "image/svg+xml":
+							case "image/webp":
 								return {
 									type: "image",
 									image: blob,
 								} as ImagePart;
+
+							case "image/svg+xml":
+								args.context.logger.warn(
+									{ id: fileParameter.id, type: fileParameter.type, name: fileParameter.name },
+									"SVG is not supported for vision input. Please upload PNG/JPEG/GIF/WebP instead.",
+								);
+								return undefined;
+								
 							case "application/pdf":
 								return {
 									type: "file",


shige (changes requested):
## Review

Thanks for fixing the SVG handling issue! The core fix is correct — returning `undefined` for SVG and adding WebP support are both good changes. However, there are a couple of things I'd like to see addressed before merging.

### 1. Trailing whitespace (nit)

**`packages/giselle/src/generations/internal/use-generation-executor.ts` line 582**

There's trailing whitespace after `return undefined;`. Could you clean that up?

### 2. UI still accepts SVG uploads for image nodes

**`internal-packages/workflow-designer-ui/src/editor/properties-panel/file-node-properties-panel/index.tsx` line 18**

The image upload accept list still includes `image/svg+xml`:

```typescript
accept: ["image/png", "image/jpeg", "image/gif", "image/svg+xml"],
```

With this PR, SVG files uploaded through the Image node will be silently dropped at generation time — the user won't see the warning log. This creates a confusing experience where the upload succeeds but the image is quietly ignored.

Could you remove `image/svg+xml` from this accept list so users can't upload SVGs as images in the first place? Also, since the backend now supports `image/webp`, it would be good to add it here too:

```typescript
accept: ["image/png", "image/jpeg", "image/gif", "image/webp"],
```

### 3. Misleading validation message for SVG in `file-panel.tsx`

**`internal-packages/workflow-designer-ui/src/editor/properties-panel/file-node-properties-panel/file-panel.tsx` line 53**

This validation returns `"Please use Image node to upload this file."` for `image/svg+xml`. Since SVGs will no longer work as image inputs, this message becomes misleading. If SVG is removed from the image accept list (as suggested above), this case can also be removed or updated to something like `"SVG is not supported. Please use PNG/JPEG/GIF/WebP instead."`.

bethany2208:
Looks like this PR has been inactive for a while. I’d be happy to pick up the remaining fixes (SVG handling in image nodes + UI updates) and open a follow-up PR if that’s okay.

shige:
Thank you for taking the time to open this pull request, and we are truly sorry for leaving it unattended for so long.

We are no longer able to actively maintain the project in a way that lets us review external contributions, so we have changed our contribution policy. You can read the details here: https://github.com/giselles-ai/giselle/pull/2941

For this reason, we are very sorry, but we will be closing this pull request. We sincerely appreciate your interest in and contribution to Giselle.

PR #2750, opened 2026-02-24
feat(protocol): add structured output protocol for End Node

WashizuRyo:
## Summary
Add protocol types to support structured JSON output from End Nodes. This introduces an `Output` discriminated union (`passthrough` | `object`) to `EndContent`, enabling End Nodes to define a JSON schema and map each schema property to a specific node output via `PropertyMapping`.

## Changes
- Add `PropertyMapping`, `EndOutput`, and `Output` types to `EndContent` in `packages/protocol/src/node/operations/end.ts`
- `Output` is a discriminated union: `passthrough` (existing behavior) or `object` (structured output with `schema` + `mappings`)
- `PropertyMapping` links a schema property path (`path`, nonempty) to a node output reference (`source`) with optional nested path access
- Mapping source is inlined into `PropertyMapping` (nodeId, outputId, path)
- Add default `output: { format: "passthrough" }` to End Node factory in `node-registry`
- Fix test in `is-supported-connection.test.ts` for the new required `output` field

## How `mappings` works

`mappings` is an array of `PropertyMapping` that resolves each schema property to an actual value at task execution time. Each mapping has two independent paths:

- **`path`** (nonempty): where to write the value in the output schema (e.g. `["report", "score"]`)
- **`source.path`**: where to read the value from the source node's output (e.g. `["metrics", "score"]`). When `[]`, the entire output is used.

Example: Text Gen Node A outputs `{ "metrics": { "score": 95 } }` and Text Gen Node B outputs `"passed"`. When the End Node has the following schema:

```json
{
  "title": "evaluation",
  "type": "object",
  "properties": {
    "report": {
      "type": "object",
      "properties": {
        "score": { "type": "number" },
        "result": { "type": "string" }
      },
      "additionalProperties": false,
      "required": ["score", "result"]
    }
  },
  "additionalProperties": false,
  "required": ["report"]
}
```

The `mappings` would be:

```json
[
  {
    "path": ["report", "score"],
    "source": { "nodeId": "nd_textgen_a", "outputId": "otp_1", "path": ["metrics", "score"] }
  },
  {
    "path": ["report", "result"],
    "source": { "nodeId": "nd_textgen_b", "outputId": "otp_2", "path": [] }
  }
]
```

At execution, the backend iterates each mapping, fetches the source node's generation output, extracts the value at `source.path` (`[]` = entire output), and writes it to the location specified by `path`. The result is assembled into `{ "report": { "score": 95, "result": "passed" } }`.

## Testing
- `pnpm build-sdk` -- pass
- `pnpm check-types` -- pass
- `pnpm test` -- 229 tests pass
- `pnpm tidy` -- clean

## Summary by CodeRabbit

* **New Features**
  * End nodes now support output configuration with multiple format options, including passthrough and structured formats with property mapping capabilities.

### DIFF
--- a/packages/protocol/src/node/operations/end.ts
+++ b/packages/protocol/src/node/operations/end.ts
@@ -2,16 +2,13 @@ import * as z from "zod/v4";
 import { Schema } from "../../structured-output";
 import { NodeId, OutputId } from "../base";
 
-const Source = z.object({
-	nodeId: NodeId.schema,
-	outputId: OutputId.schema,
-	path: z.array(z.string()),
-});
-export type Source = z.infer<typeof Source>;
-
 const PropertyMapping = z.object({
-	path: z.array(z.string()),
-	source: Source,
+	path: z.array(z.string()).nonempty(),
+	source: z.object({
+		nodeId: NodeId.schema,
+		outputId: OutputId.schema,
+		path: z.array(z.string()),
+	}),
 });
 export type PropertyMapping = z.infer<typeof PropertyMapping>;
 


WashizuRyo:
Done ✅ https://github.com/giselles-ai/giselle/pull/2750#issuecomment-3948652066

shige (approved):
Thank you! 🚀

